PR #1, opened 2022-02-18
Nickel shell

yannham:


README.md
```diff
@@ -1,2 +1,19 @@
 # nickel-nix
-An experimental Nix toolkit to use nickel as a language for writing nix packages, shells and more.
+
+An experimental Nix toolkit to use nickel as a language for writing nix
+packages, shells and more.
+
```
garbas, line 5: We need to mention that only equivalent to `nix-shell` is provided, but more is to come. This is to manage expectations when somebody comes to here.

garbas (approved):

examples/nix-shell/shell.ncl
```diff
@@ -0,0 +1,20 @@
+{
+  inputs = {
+    hello = { input = "nixpkgs" }
+  }
+
+  , output = fun { pkgs, nix, .. } =>
```
thufschmitt, line 6: I think it would be nice to make it explicit that this `pkgs` is derived from the `inputs` field. I remember we didn’t want to call it `inputs` because it could lead to confusing errors since there’s already an `inputs` in scope, but maybe we could rename the `inputs` field to `inputs_spec` or something similar to prevent this.

(besides, `pkgs` isn’t restricted to packages, so it would be nice for it to have a broader name)

examples/nix-shell/shell.ncl
```diff
@@ -0,0 +1,20 @@
...
+    { # (optional) name of the shell
```
thufschmitt, line 7: It’s a bit strange to have one (slightly misindented) comment for this field and none for the others. Maybe we could comment all or none of them.

Also, shouldn’t we use annotations for this kind of stuff?

lib.nix
```diff
@@ -0,0 +1,131 @@
...
+  exportForNickel = value:
+      let type = builtins.typeOf value; in
+      if (type == "set") then (
+        if (value.type or "" == "derivation") then
+          { type = "nixDerivation"; drvPath = value.drvPath; outputName = value.outputName; }
+        else
+          builtins.mapAttrs (_: exportForNickel) value
+        )
+      else if (type == "list") then builtins.map exportForNickel value
+      else if (type == "list") then
```
thufschmitt, line 12:
```suggestion
      else if (type == "lambda") then
```

lib.nix
```diff
@@ -0,0 +1,131 @@
...
+        builtins.map importFromNickel_ value
+    else value;
+
+  # Generate a Nickel program that evaluates the nickel-nix output, passing
+  # the given exported packages, and write it to outFile.
+  computeNickelFile = system: {nickelFile, exportedPkgs}:
+    let
+      exportedJSON = builtins.toFile
+          "inputs.json"
+          (builtins.unsafeDiscardStringContext (builtins.toJSON (exportForNickel exportedPkgs)));
```
thufschmitt, line 46: This `unsafeDiscardStringContext` is mildly nasty. It would be cleaner to bring `nixpkgs` into scope and use `pkgs.writeText` (which allows bringing in a context) rather than `builtins.toFile`. But I guess depending on nixpkgs is annoying, so feel free to ignore me :wink:

lib.nix
```diff
@@ -0,0 +1,131 @@
...
+      pkgNames = builtins.attrNames declaredInputs;
+      addPackage = name: acc:
+        let inputName = declaredInputs."${name}".input; in
+        if builtins.hasAttr inputName flakeInputs then
+          let
+            input =
+              if inputName == "nixpkgs" then
+                flakeInputs.${inputName}.legacyPackages.${system}
+              else
+                flakeInputs.${inputName}.${system}.packages;
```
thufschmitt, line 89:
```suggestion
                flakeInputs.${inputName}.packages.${system};
```

thufschmitt: Ideally I think this whole magic doesn’t really belong here (it should be up to the user to specify the exact attribute path, like in plain Nix). But we can keep it for a first POC

thufschmitt (changes requested):
Nice :sunglasses: 

A number of small comments, but very cool overall.

My big gripe right now is that it didn’t work out of the box, I had to update the lockfiles to prevent some weird nickel error:

```
nickel-res.json> error: Type error
nickel-res.json>     ┌─ /nix/store/d8kj952i389kdsiwfjiwzyrf79idmmz7-shell.ncl:14:22
nickel-res.json>     │
nickel-res.json>  14 │        , shellHook = nix.lib.nix_string_hack [
nickel-res.json>     │                      ^^^^^^^ This expression has type <unevaluated>, but Record was expected
```

But with some up-to-date lockfiles, everything went swell

thufschmitt:
Fwiw, I wrote a version at https://github.com/thufschmitt/nickel-nix/tree/nicke-shell-pure that doesn’t require `--impure`. But it makes the implementation absolutely ugly, so I don’t think it’s worth merging here.

examples/nix-shell/shell.ncl
yannham: I agree with the strange indentation.

I think those are more like comments to explain the example further, but not documentation that is meant to be queried (in fact, `nix.ncl` may already define a documentation for those fields, or is likely to do so in the future).

lib.nix
yannham: To be honest I just copy pasted to other code that exports thing. We already depend on a bunch of things, so it's fine to have one one more dependency. Could we just depend on `writeText` directly?

lib.nix
yannham: I kinda agree, but for the POC we didn't want to have to specify flake-related stuff that doesn't appear in the Nix version of the shell.

yannham:
> My big gripe right now is that it didn’t work out of the box, I had to update the lockfiles to prevent some weird nickel error:

I suspect the problem is that I wanted to make this PR such that it would work directly once merged into master, so the flake input points to master which is empty currently. Still, there the lockfiles? I don't know, we may have to do an update after we merge this PR.

> Fwiw, I wrote a version at https://github.com/thufschmitt/nickel-nix/tree/nicke-shell-pure that doesn’t require --impure. But it makes the implementation absolutely ugly, so I don’t think it’s worth merging here.

Nice, I'm curious about it, I'll take a look at some point.

examples/nix-shell/shell.ncl
```diff
@@ -0,0 +1,20 @@
...
+
+       , packages = [
+         pkgs.hello
+       ]
+
+       , shellHook = nix.lib.nix_string_hack [
+          "echo 'Development shell'",
+          pkgs.hello,
+          "/bin/hello"
+       ]
```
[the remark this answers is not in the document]
yannham: Actually it can't work with cowsay. `lib.nix` used the declared inputs to pass to Nickel whatever is needed, because JSON is used as an exchange format, everything that is serialized must be forced. So we have to know what to export to JSON in advance. I guess this makes nix_string_hack totally useless in our model?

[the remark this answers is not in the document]
yannham: Then I don't understand how it's different from `"hello"` or useful. If I put it into the inputs, can't I just use a native Nickel string:

```
shellHook = m%"
  %{pkgs.cowsay}/bin/cowsay 'Welcome to this nickel-powered shell'
 "%m
```

?

lib.nix
yannham: I just tried and it indeed starts to bring in a lot of things, it seems. I guess the `stdenv` and all?

thufschmitt: Yes, you can’t really get it without pulling the whole world. So `toFile` will be enough for the time being :))

examples/nix-shell/shell.ncl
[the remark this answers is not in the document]
thufschmitt: Technically I think nickel wouldn’t evaluate that `pkgs.cowsay` _probably_ can’t be directly interpolated to a string (but that’s a detail and we could change that).

But the more fundamental reason preventing this is that having a variable in `inputs` doesn’t mean that it will be an input of every derivation in the file − and it’s actually not added as input unless you explicitely specify it.

As a (really contrieved example), https://github.com/thufschmitt/nickel-nix/tree/compound-shell has an intermediate derivation, and the final shell depends on `perl`, but not the intermediate derivation

examples/nix-shell/shell.ncl
yannham: Thought hard about a better name, `inputs_decl`, `inputs_requested`, etc. but went with `inputs_spec` in the end. Not great, but the less badest :slightly_smiling_face:

yannham:
Few, I think we're done. Tested again, it seems to work. I'm gonna merge this and test if everything is fine with inputs pointing to `main`, and potentially fix issues in separate PRs.

PR #6, opened 2022-03-12
fix addPackage if inputName != "nixpkgs"

bbigras:
I have no idea if it's right.

I needed this change with `github:serokell/deploy-rs` but maybe `packages` and `system` is inverted on their side.


yannham:
Hello,

It is right indeed, and was actually suggested during the review: https://github.com/nickel-lang/nickel-nix/pull/1#discussion_r822447477. Must have been lost in the fury of pre-release times :slightly_smiling_face: 

Thanks you!

yannham (approved):

PR #8, opened 2022-03-25
Improve both READMEs

yannham:
- Add a word of caution about the scope of the project
- Put more directly how to try the lib in the main README
- Add a section on why one wants to use Nickel for Nix


yannham:
In case this escaped your radar @aspiwack , a gentle ping :slightly_smiling_face:

README.md
```diff
@@ -18,7 +26,137 @@ The Nickel library contains in-code documentation that can be leveraged by the
...
+user experience for Nix[^1]. In particular, Nickel adds validations capabilities
+thanks to its type system and contract system. As a simple illustration, let's
+make a stupid error in our Nix version of the `hello` shell,
+[`examples/nix-shell/shell.nix`](examples/nix-shell/shell.nix), by replacing the
+list of `packages` by a string:
+
+```nix
+-    packages = [ pkgs.hello ];
++    packages = "pkgs.hello";
+```
```
aspiwack, line 61: I think it would be better to provide attributes to the flake which do use such faulty implementations. Instead of asking people to make changes. What do you think?

yannham: Yup, good idea.

PR #10, opened 2022-10-18
Use modular builders to build a simple C hello program

yannham:
This PR changes the structure of the nickel-nix to be a starting point to write modular and composable builders, in order to create packages or devshell. We don't want to rely anymore on `mkShell` on the Nix side but use pure Nickel abstraction over the simple `derivation` function.

## Content

This PR adds the most basic builder (a bit of sugar over `derivation`), and clean many parts of the Nickel library. The shell example has been removed, and a purely `derivation`-based example shell will be added in a subsequent PR. We also get rid of `nix_string_hack` and use nickel's very new symbolic strings to emulate Nix string contexts.

## Next

- should we get rid of the `sources` input type, now that we have symbolic strings and `import_file`? (If yes, let's do this in another PR)
- add a basic shell builder from the `devshell`/`nickel-devshell` branch


examples/c-hello-world-string-ctxt/README.md
```diff
@@ -0,0 +1,29 @@
+# C hello world
+
+This example builds a simple C hello wolrd program through nickel-nix, using
```
rapenne-s, line 3:
```suggestion
This example builds a simple C hello world program through nickel-nix, using
```

examples/c-hello-world/README.md
```diff
@@ -0,0 +1,37 @@
+# C hello world
+
+This example builds a simple C hello wolrd program through nickel-nix.
```
rapenne-s, line 3:
```suggestion
This example builds a simple C hello world program through nickel-nix.
```

rapenne-s (commented):
LGTM

rapenne-s (approved):

yannham:
@thufschmitt @djacu gentle reminder to review this PR :)

examples/c-hello-world/flake.nix
```diff
@@ -0,0 +1,26 @@
+{
+  description = "A basic flake with a shell";
+
+  inputs.nixpkgs.url = "nixpkgs/nixos-unstable";
+  inputs.flake-utils.url = "github:numtide/flake-utils";
+  # inputs.nickel-nix.url = "github:nickel-lang/nickel-nix";
+  inputs.nickel-nix.url = "path:../../";
```
djacu, line 7: Using "git+file:../../" here seems to fix a `error: cannot fetch input 'path:../../?lastModified=0&narHash=sha256-12fmSIyeZCB5XWAFSIm4AYYVlvI5eQLeg2FXPcbaMsc=' because it uses a relative path`

See https://github.com/NixOS/nix/issues/3978#issuecomment-907573179

Still cannot use things like `nix flake check` and `nix flake show` but seems to get us a little closer.

djacu: Actually, I get the same error when trying to use `nix build .#` but it goes away when using `git+file` so this appears to have more benefit.

yannham: Oh, good catch. In fact we set the relative path for development only: in this PR, I think we can just get rid of it and restore the commented input.

yannham: I think I'll have to do this in two steps. I first fixed the input, but for it to work I need this repo to have the latest version of Nixel, i.e. to merge this PR first. I'll update the inputs of the example in a second step.

PR #20, opened 2023-01-03
fix examples to print a newline

rapenne-s:
depending on your shell, you may have a symbol displaying the lack of newline, or have your prompt after the message which may be confusing and could lead the reader to think nothing was displayed

I originally planned to add more fixes, but it turned it's a nix problem, so not much here


yannham:
The target branch is not master: is this expected, @rapenne-s ?

rapenne-s:
yes, the examples don't exist in branch `main`, it's still in the working branch `dev-shell/bash`

yannham (approved):

PR #21, opened 2023-01-16
Devshell builders: bash

yannham:
Make the simplest builder possible: a devshell opening bash. There is a fair bit of dirty things here (we copy the ncl files from the library directly into the example, we move the whole flake dir in the store, etc.) but we could at lest make it work.


yannham:
cc  @djacu

vkleen:
Yesterday's experiments are at #22

yannham:
Merged in `/future`.

PR #22, opened 2023-01-19
We did something!

vkleen:
Extremely hacky prototype for structured_env; it should probably be a record of records instead of a record of arrays.


rapenne-s:
do we need to keep it?

vkleen:
We only really added the `structured_env` handling. I think that's already in `main` now after our rebase spree. So, if that's correct, then this can be closed.

yannham:
Superseded by #29.

PR #23, opened 2023-01-26
We did something again!

djacu:


rapenne-s:
I added 10 builders to this branch in the same file where the Rust builder is.

However, Javascript and PHP builders aren't working because their LSP is marked as broken in nixpkgs, and for some reasons I have an error with the Python310 environment.

rapenne-s:
I added a script create directories in the `example` directory for each development environment supported in the file `builders.ncl` of the directory `rust-dev-shell`

PR #26, opened 2023-01-26
updated flake lock and added devshell

djacu:


thufschmitt:
(I don't have merge access it seems. @yannham or other, care to merge this?)

PR #29, opened 2023-01-30
Task/merge future

yannham:
It seems like #10 wasn't merge into `main`, but into `future`. This PR rebases the current state of `future` onto main, and merge. There shouldn't be much difference with #10, but a quick review is still advised.


yannham:
It might have butchered the README. But I think we can rework this once we have merged the devshell builders, because it'll be quite different eventually (the current README heavily refers to the previous shell example, but this PR gets rid of it).

rapenne-s (approved):
LGTM

djacu (approved):
lgtm!

PR #31, opened 2023-02-01
Make example templates

djacu:
Moved the rust devshell to a templates directory and made it work with Nix templates.

From the rust template directory you can enter the devshell using.
```
nix develop --impure --override-input nickel-nix ../../
```

If you want to test out the template, from an empty directory run
```
nix flake init --template <path-to-nickel-nix>/templates#rust-devshell
```
and then 
```
nix develop --impure --override-input nickel-nix <path-to-nickel-nix>
```

Once this is merged, users can do the same with
```
nix flake init --template github:nickel-lang/nickel-nix#rust-devshell
```
and
```
nix develop --impure
```
respectively.


rapenne-s (approved):
looks good, thank you

PR #32, opened 2023-02-01
Add more devshell templates

djacu:
Adds the rest of the devshells that @rapenne-s added in nickel as accessible devshell templates.


rapenne-s:
Javascript and PHP aren't working (this is known) because of a broken package in nixpkgs, however, now when using `nix develop`, we have a nix error instead of the broken package message.

Python310 didn't work for some reasons, and same as above, there is a nix error in lib.nix that we didn't have before.

```
error: attribute 'input' missing                                                                                                                                                                                                                                                                                                                                                                                                      
       at /nix/store/5xzbdraf0d7ahgnrs1nzi65lyjxg2f3r-source/lib.nix:115:25:                                                                                                                                                                                                                                                                                                                                                                    114|       addPackage = name: acc:
          115|         let inputName = declaredInputs.${name}.input; in
             |                         ^                                                                                                                                                                                     116|         # "sources" is a special type of input for files. They mimic Nix style
(use '--show-trace' to show detailed location information)    
```

I think we shouldn't provide these templates given the current state, they don't work and users wont' have a message explaining why, they will think nickel-nix is broken.

djacu:
I added some logic to the templates generator so it excludes the broken shells.

```shell
$ nix flake show
git+file:///home/bakerdn/tweag/nickel-nix?ref=refs%2fheads%2fadd-more-devshell-templates&rev=2352ae77888c65bce10d1c249ab759de72fd8e12
├───devShells
│   ├───aarch64-darwin
│   │   └───default: development environment 'nix-shell'
│   ├───aarch64-linux
│   │   └───default: development environment 'nix-shell'
│   ├───x86_64-darwin
│   │   └───default: development environment 'nix-shell'
│   └───x86_64-linux
│       └───default: development environment 'nix-shell'
├───lib: unknown
└───templates
    ├───c-devshell: template: A c devshell using nickel.
    ├───clojure-devshell: template: A clojure devshell using nickel.
    ├───erlang-devshell: template: A erlang devshell using nickel.
    ├───go-devshell: template: A go devshell using nickel.
    ├───racket-devshell: template: A racket devshell using nickel.
    ├───rust-devshell: template: A rust devshell using nickel.
    ├───scala-devshell: template: A scala devshell using nickel.
    └───zig-devshell: template: A zig devshell using nickel.
```

rapenne-s:
we can celebrate this 🍾🍾

PR #34, opened 2023-02-03
Replaces structured_env array for record

djacu:
Changes the structured_env so that its elements use records instead of arrays so they are more composable.

### DIFF
--- a/templates/devshells/c/builders.ncl
+++ b/templates/devshells/c/builders.ncl
@@ -74,15 +74,15 @@ let {NickelDerivation, ..} = import "contracts.ncl" in
     },
     inputs,
     output.structured_env = {
-      PATH = [
-        s%"%{inputs.bash}/bin"%,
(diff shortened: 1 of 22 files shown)


thufschmitt (approved):
Looks good :+1:  (except that it's gonna conflict with #33 , not sure which one we want to merge first)

yannham (approved):
LGTM in general, but I wonder if the `bash = ` shouldn't go in the base bash shell instead of being repeated in each language-specific shell.

Otherwise, I think that if you try to combine a C shell and a Rust shell today, you might get into trouble because of the non idempotency of merging. Well, because those are derivation and strings, it might actually work as it is, but at least you don't have to repeat it in each and very builder. And it might break later if we change e.g. the representation of derivations.

yannham:
This one works, so let's merge it now. There way too many PR in the queue.

PR #38, opened 2023-02-04
Add CI

djacu:
Adds CI for testing the devshell templates.

Okay I have good news and bad news.
Good news: It works!
Bad news: It takes a while...

### DIFF
--- a/.github/workflows/templates.yml
+++ b/.github/workflows/templates.yml
@@ -28,9 +28,9 @@ jobs:
     - name: List all templates
       id: templates
       env:
-      # Should only be needed until we use a `cachinx/install-nix` action that contains
-      # https://github.com/cachix/install-nix-action/pull/152 (should be in `v19`)
-      NIX_CONFIG: "extra-access-tokens = github.com=${{ secrets.GITHUB_TOKEN }}"
+        # Should only be needed until we use a `cachinx/install-nix` action that contains
+        # https://github.com/cachix/install-nix-action/pull/152 (should be in `v19`)
+        NIX_CONFIG: "extra-access-tokens = github.com=${{ secrets.GITHUB_TOKEN }}"
       run: |
         templates=$(nix eval .#templates --apply builtins.attrNames --json)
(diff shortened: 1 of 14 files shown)


rapenne-s:
maybe it would be a lot faster if you use nickel cachix substituter? Because otherwise you have to compile nickel every time, which is wrong

djacu:
@rapenne-s you were totally right. It's much faster now with the latest commit.

templates/devshells/c/flake.nix
```diff
@@ -5,6 +5,11 @@
   inputs.flake-utils.url = "github:numtide/flake-utils";
   inputs.nickel-nix.url = "github:nickel-lang/nickel-nix";
 
+  nixConfig = {
+    extra-substituters = ["https://tweag-nickel.cachix.org"];
+    extra-trusted-public-keys = ["tweag-nickel.cachix.org-1:GIthuiK4LRgnW64ALYEoioVUQBWs0jexyoYVeLDBwRA="];
+  };
+
```
rapenne-s, line 12: This comes from the template directory generating the templates directories, you should commit the original file too

.github/workflows/templates.yml
```diff
@@ -0,0 +1,67 @@
...
+
+    - name: Initialize devshell ${{ matrix.template }}
+      run: |
+        nix flake new --template path:$PWD/repo#${{ matrix.template }} example --accept-flake-config
+
+    - name: Enter devshell ${{ matrix.template }}
+      run: |
+          pushd ./example
+          nix develop --impure --accept-flake-config
+          popd
```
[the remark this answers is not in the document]
djacu: I believe it is working. The CI used to take 10-15 minutes before adding the substituters, and now it takes 1 minute. From looking at the CI logs, it appears to do what I think it should to.

templates/devshells/c/flake.nix
rapenne-s: nevermind, I thought we still had a `rust-devshell` directory where development is happening, and the `rust` example that was created from it :sweat_smile:

PR #39, opened 2023-02-08
Haskell devshell template

andreabedini:
I basically copied the rust template and changed the inputs. Do we need to keep all `builders.ncl` in sync?
Let me know if I am doing it wrong.

### DIFF
--- a/templates/devshells/haskell/builders.ncl
+++ b/templates/devshells/haskell/builders.ncl
@@ -1,0 +1,281 @@
+let {NickelDerivation, ..} = import "contracts.ncl" in
+
+{
+  NickelPkg
+    # we should only need two '%%', but a current Nickel bug (#XXX) bug makes the example being
+    # recognized as actual interpolation. For the time being, we thus use a
+    # three '%' delimiter.
+    | doc m%%%"
+      The most basic builder. Simply alias to `NickelDerivation`.
+
+      Example:
+
+      Here is a simple example of a function producing a Nickel derivation
+      using the `NickelPkg` builder.
+
+      ```nickel
+      output = fun { inputs, nix, .. } =>
+      {
+        name = "hello",
+        version = "0.1",
+        build_command = {
+          cmd = s%"%{inputs.bash}/bin/bash"%,
+          args = [ "-c",
+            s%"
+              %{inputs.gcc}/bin/gcc %{nix.lib.import_file "hello.c"} -o hello
+              %{inputs.coreutils}/bin/mkdir -p $out/bin
+              %{inputs.coreutils}/bin/cp hello $out/bin/hello
+             "%
+          ]
+        },
+        env = {},
+      } | nix.builders.NickelPkg
+      ```
+      "%%%
+    = NickelDerivation,
+
+  BashShell = {
+    inputs_spec
+      | {
+        # default or not?
+        bash.input | priority -100 = "nixpkgs",
+        nakedStdenv.input | priority -100 = "myInputs",
+        ..
+      }
+      | default = {},
+
+    inputs,
+    nix,
+
+    output = {
+      version | default = "0.1",
+
+      # this is required otherwise nix develop
+      # will fail with a message that it only supports bash
+      build_command = {
+        cmd = s%"%{inputs.bash}/bin/bash"%,
+        args = [],
+      },
+
+      env = {
+        # TODO: handle naked derivations without having to interpolate
+        stdenv = s%"%{inputs.nakedStdenv}"%,
+      },
+    } | NickelPkg,
+  },
+
+  RustShell = BashShell & {
+    inputs_spec | {
+      cargo.input | priority -100 = "nixpkgs",
+      rustc.input | priority -100 = "nixpkgs",
+      rustfmt.input | priority -100 = "nixpkgs",
+      rust-analyzer.input | priority -100 = "nixpkgs",
+      ..
+    },
+    inputs,
+    output.structured_env = {
+      PATH = [
+        s%"%{inputs.bash}/bin"%,
+        s%"%{inputs.cargo}/bin"%,
+        s%"%{inputs.rustc}/bin"%,
+        s%"%{inputs.rustfmt}/bin"%,
+        s%"%{inputs.rust-analyzer}/bin"%
+      ],
+    },
+    output.env = record.map (fun _n xs => array.foldl (fun acc x => s%"%{acc}:%{x}"%) "" xs) output.structured_env,
+  },
+
+  GoShell = BashShell & {
+    inputs_spec | {
+      go.input | priority -100 = "nixpkgs",
+      gopls.input | priority -100 = "nixpkgs",
+      ..
+    },
+    inputs,
+    output.structured_env = {
+      PATH = [
+        s%"%{inputs.bash}/bin"%,
+        s%"%{inputs.go}/bin"%,
+        s%"%{inputs.gopls}/bin"%,
+      ],
+    },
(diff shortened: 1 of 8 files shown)


templates/devshells/haskell/builders.ncl
```diff
@@ -0,0 +1,281 @@
...
+    },
+    output.env = record.map (fun _n xs => array.foldl (fun acc x => s%"%{acc}:%{x}"%) "" xs) output.structured_env,
+  },
+
+  HaskellShell = BashShell & {
+    inputs_spec | {
+      cabal-install.input | priority -100 = "nixpkgs",
+      ghc.input | priority -100 = "nixpkgs",
+      haskell-language-server.input | priority -100 = "nixpkgs",
+      stack.input | priority -100 = "nixpkgs",
```
aspiwack, line 265: Why is the priority `-100` and not `default`?

aspiwack (commented):
I'm not completely sure why a Haskell template contains shells for Erlang and Clojure.

That being said, I'd say this isn't a particularly good default for a Haskell dev shell. First, you probably want to use [Julien's method](https://www.tweag.io/blog/2022-06-02-haskell-stack-nix-shell/) for Stack, in order for it to work both on Nixos and non-Nixos. You'll also need the ability to choose the GHC version that you use to coordinate with your Stack snapshot (I'm actually not too sure how to retrieve the default version from Nixpkgs, so that's a good question). You will need `nix` in the shell as well, because Stack's Nix integration will want it (maybe that's built in, though?)

I came up with a basic shell for Haskell recently, and it looks [like this](https://github.com/aspiwack/haskell-benign/blob/99bc379ac14177cb5345c6fc96d13b2996163f72/shell.nix) (you can ignore `just`, and probably `git`). It is not perfect. For once it doesn't guarantee that the installed Ormolu version and the version linked in the HLS is the same (maybe I should use the direct invocation of HLS, rather than the provided wrapper, I'm not sure what the difference is). In fact, maybe we shouldn't tie the Ormolu version with the GHC version at all! (also the devtool bit is to avoid downloading HLS on CI because it's more than 1GB of closure… but this should be trivial to achieve in Nickel shells, hopefully).

All that to say: getting Haskell right is tricky. I'm not against merging this minimalistic proposal, but I strongly suggest we don't leave it at that, because it's not going to be very useful to many users.

templates/devshells/haskell/builders.ncl
yannham: I _think_ when we first made the shells we had a stupid issue like the basic `NickelPackage` contract defines a default somewhere and then we have two conflicting default, so we had to make the priority "one layer up, but still easily overridable". I would say that this is orthogonal to the Haskell shell in particular, but in parallel we will definitely revisit this to see what is the issue again and if we can get rid of those pesky `-100`.

rapenne-s: @aspiwack the file `builders.ncl` is the library containing all the builders, then in your nickel file you can merge all the builders you want to create your shell.

aspiwack: @rapenne-s is `builders.ncl` copied in every template then?

andreabedini: @aspiwack Getting haskell right is hard that's true (... in general, with or without nix! :P). Happy to implement Julien's  wrapper. Regarding the GHC version, the pragmatic approach would be to allow the user to choose what version in a simple way. Then it's up to the user making sure the choosen LTS matches the GHC version (if they use stack).

Would this work for you?

aspiwack: As soon as the ghc version is overridable, then it's easy to make sure that everything is compatible in the style of my Nix shell above (well, except the Stackage snapshot, but I think this is pretty unavoidable).

templates/devshells/haskell/flake.nix
```diff
@@ -0,0 +1,39 @@
+{
```
sir4ur0n, line 1: > I came up with a basic shell for Haskell recently, and it looks [like this](https://github.com/aspiwack/haskell-benign/blob/99bc379ac14177cb5345c6fc96d13b2996163f72/shell.nix) (you can ignore just, and probably git). It is not perfect. For once it doesn't guarantee that the installed Ormolu version and the version linked in the HLS is the same (maybe I should use the direct invocation of HLS, rather than the provided wrapper, I'm not sure what the difference is). In fact, maybe we shouldn't tie the Ormolu version with the GHC version at all! (also the devtool bit is to avoid downloading HLS on CI because it's more than 1GB of closure… but this should be trivial to achieve in Nickel shells, hopefully).

@aspiwack (and others if you are interested) for what it's worth, in Zeus we have done the following: https://github.com/tweag/zeuslogics/blob/master/nix/default.nix#L62-L76
i.e. use an overlay to retrieve the Ormolu package used by HLS (otherwise we had inconsistent formatting between formatting with HLS and formatting in the CL/`pre-commit-hooks.nix`).

Courtesy of @amesgen for the trick

templates/devshells/haskell/builders.ncl
rapenne-s: @aspiwack yes, to be ready for FOSDEM we made things easier, we have the rust example containing a script creating all the other templates using the rust as a template itself.

We certainly want something better in the future, but it's still not clear how we want to distribute Nickel-nix to users (or its installation process).

templates/devshells/haskell/flake.nix
aspiwack: This hurts my brain a little. But thanks for the trick! This sounds quite useful.

templates/devshells/haskell/builders.ncl
aspiwack: Understood.

TaylorEwtonQ:
This is a discussion regarding the Haskell devshell template proposed by andreabedini. Aspiwack reviewed the proposal and suggested that it is not a good default template for Haskell development. Aspiwack suggests using Julien's method for Stack, which works for both Nixos and non-Nixos. Aspiwack also recommends using nix in the shell because Stack's Nix integration requires it. Aspiwack provided a basic shell for Haskell that looks like the following:

thufschmitt:
Closing in favour of #67

templates/devshells/haskell/flake.nix
thufschmitt: > it doesn't guarantee that the installed Ormolu version and the version linked in the HLS is the same

In which situation can this happen? My understanding was that each `haskellPackages` set had only one version for each package (barring custom overrides), so `haskell.packages.{blah}.ormolu` and `haskell.packages.{blah}.haskell-language-server` should use the same Ormolu version, right?

thufschmitt: Aaah indeed, so “barring custom overrides” is not just a theoretical concern. Thanks for the pointer.

> Partially related, starting in [HLS 2.3.0.0](https://github.com/haskell/haskell-language-server/releases/tag/2.3.0.0), I added the possibility to configure HLS to use an external ormolu binary via a configuration flag. For more information, see https://github.com/haskell/haskell-language-server/pull/3771

I've seen that, yes! Hopefully we'll just use that in the future (but the default HLS in Nixpkgs is still 1.something, so it's a bit early unfortunately)

sir4ur0n: > but the default HLS in Nixpkgs is still 1.something, so it's a bit early unfortunately

Which `nixpkgs` branch are you referring to? On `nixos-unstable` the current version is 2.3.0.0+, that's what we use for Zeus.

PR #40, opened 2023-02-09
Support path in inputs

### LINKED ISSUE #35: Make it possible to depend on a package that's not at toplevel on nixpkgs
I wanted to add a dependency to `libcurl` in a C shell. However, that required adding `curl.dev` to my inputs, which currently isn't possible since the input must have the form `inputs.${inputName}.(legacy)Packages.${blah}`. (I worked my way around by editing the Nix code to expose it as `curl_dev`, but that's very hacky)

It would be important to allow accessing nested fields inside an input, maybe by adding an optional `attribute_path | Array String` attribute to the elements of `input_spec`

yannham:
Closes #35. Introduce a `path` attribute in the `input_specs` schema, which can point to a composed path inside the corresponding input.

Additionally, this PR decouples the field name used in `input_specs` from the actual package, allowing to do something like:

```nickel
input_specs.my_local_gcc_alias = {
  input = "nixpkgs",
  path = "gcc",
}
```

Allowing several, independent `gcc` inputs to coexist.

This PR fixes the shebang of the `create-env.sh` scripts as well, which was using `sh` instead of `bash`.

The change has been tested, both with simple and composed paths. It turns out we didn't correctly applied the `NickelInputSpec` contract when extracting the inputs :sweat_smile:

### DIFF
--- a/contracts.ncl
+++ b/contracts.ncl
@@ -272,10 +272,43 @@ from the Nix world) or a derivation defined in Nickel.
     nix | {..},
   },
 
+  InputPath
+    | doc m%"
+        The path of a package in an input (usually nixpkgs). Can be either a
+        single attribute, represented as a string, or an array of string,
+        representing a composed path.
+
+        If a single attribute is provided (a `Str`), this contract normalizes the
+        result by splitting it around dots to obtain an array:
+
+        # Examples
+
+        ```nickel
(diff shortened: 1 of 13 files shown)


rapenne-s (approved):
looks good to us (@thufschmitt , @djacu and me)

PR #41, opened 2023-02-16
Update to latest symbolic string syntax

yannham:
Depends on #40

Update to latest Nickel, which uses the new symbolic strings with prefix syntax. As often with templates, a lot of changes are duplicated, so you can review only Rust's contracts.ncl/builders.ncl


yannham:
@djacu I'm not sure what happened or if the force push was on purpose, but if you want me to restore this branch to its previous state, tell me :+1:

PR #46, opened 2023-02-19
Restore correct environment

yannham:
A previous PR (#45) made the usage and structure of `structured_env` simpler. However, the effects wasn't the expected one, as the `PATH` became unset in the development shell. It seems the right fix would be to get rid of `env` altogether in the builders, and delegate environment building to `NickelDerivation`.

Unfortunately, `NickelDerivation` ends up being applied several times, and it's not idempotent (Nickel currently refuses to merge arrays together, hence the computed value for `env` fails to merge with itself).

We fix the previous bug, and move the computation of `env` to `BashShell` in the meantime (waiting for Nickel to be smarter about merging identical values).

### DIFF
--- a/builders.ncl
+++ b/builders.ncl
@@ -39,7 +39,7 @@ let {NickelDerivation, ..} = import "contracts.ncl" in
       | {
         # default or not?
         bash.input | priority -100 = "nixpkgs",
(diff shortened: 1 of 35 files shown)


djacu (approved):
LGTM

PR #49, opened 2023-02-22
Add CI for examples

djacu:

PR #51, opened 2023-03-02
Move nixel-specific files in a dedicated subdirectory

### LINKED ISSUE #47: Clean up the source file tree
Initiating a flake from a Nixel template create a lot of files, most of them being Nixel library files that the user doesn't really care about. We should put them under a subdirectory (`./nixel` or whatever)

yannham:
Depends on #49 

Closes #47. Move Nixel-specific files in a `nixel` subdirectory in both examples and template, to make the source path cleaner and focus on files that the user should actually care about.


thufschmitt (approved):
Nice :)

I was expecting this would have caused more changes to the Nickel files, it's great to see that it actually doesn't :muscle:

djacu (approved):
LGTM

PR #44, opened 2023-02-16
Clean up builders input_specs

vkleen:

PR #52, opened 2023-03-03
Clean builder input specs, episode II

yannham:
Port of #44 on latest master (seemed simple to do it again than to rebase, given the recent changes). As #44, this PR removes the useless usage of a contract and numeral priority for the inputs_spec of builders, going back to simple default value.

Passing by, I think the development shells of Python and Typescript were not doing what we expected, because of the presence of package that aren't at the toplevel of Nixpkgs (see #35). I suspect that, thanks to default value and open contracts, the previous situation was that the inputs succeed, but would just pick the top-level value (say `python310` instead of the more specific `python-lsp-server` package). The PR fixes that as well by using the `path` field introduced in #40.

Supersede #44.


yannham:
As discussed with @rapenne-s, the CI failure seems to be related to one of the Erlang package being broken in Nixpkgs, and doesn't have anything to do with this PR nor Nixel. Let's ignore it for now.

PR #59, opened 2023-04-24
Switch to latest Nickel

YorikSar:
Incorporate stdlib changes and update `nickel` input.

### DIFF
--- a/examples/c-hello-world/nixel/builders.ncl
+++ b/examples/c-hello-world/nixel/builders.ncl
@@ -68,13 +68,13 @@ let {NickelDerivation, ..} = import "contracts.ncl" in
       # `env` will be automatically up-to-date
       env =
         structured_env
-        |> record.map (fun _n xs =>
+        |> std.record.map (fun _n xs =>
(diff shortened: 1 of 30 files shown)


YorikSar:
Note that `examples/c-hello-world` will require a flake lock bump after we land this since it references nickel-nix while being its part.

examples/c-hello-world/nixel/builders.ncl
```diff
@@ -90,7 +90,6 @@ let {NickelDerivation, ..} = import "contracts.ncl" in
 
     output.structured_env = {
       PATH = {
-        bash = nix-s%"%{inputs.bash}/bin"%,
```
yannham, line 93: This change doesn't seem to be related to the Nickel version. What is the reason to remove bash?

yannham (approved):

examples/c-hello-world/nixel/builders.ncl
YorikSar: This file was out of sync with one in the root. I’ve just copied the file from root here.

PR #63, opened 2023-05-09
Import nickel-nix parts from nickel-nix input

### LINKED ISSUE #24: Don't copy the library at each use site
The current way of using Nixel is to copy the Nickel part of the library into each repository that uses it. That's quite bad, even as a temporary solution because it makes it really tough to keep track of the different versions that are used (even inside the examples repository, there's already a number of different copies).

A simple but better solution could be to still vendor it, but either as a git submodule (for external projects) or as a symlink (for the in-repo examples) so that things stay internally coherent internally, and it's easier to track the versions externally.

YorikSar:
Move everything from `nixel` repo in templates to the `nickel-nix` repo itself and import from it as input in the templates.

This will require user to run `nix run .#generate-lockfile` after flake initialisation and on each change of the flake lock file.

Fixes #24


templates/devshells/c/flake.nix
```diff
@@ -35,5 +34,13 @@
       devShells = {
         default = nickelDerivation;
       };
+      apps = {
+        regenerate-lockfile = {
+          type = "app";
+          program = pkgs.lib.getExe (buildLockFile {
```
thufschmitt, line 40: I'm not sure why, but this `getExe` seems to be the thing causing the CI failure.

thufschmitt: Gne, nevermind that one, it's Friday

thufschmitt: The actuall issue is

```suggestion
          program = pkgs.lib.getExe (nickel-nix.lib.${system}.buildLockFile {
```

YorikSar: It's already imported in let block above, so this shouldn't be the issue...

templates/devshells/c/flake.nix
```diff
@@ -35,5 +34,13 @@
...
+            nickel-nix = nickel-nix.lib.${system}.lockFileContents;
+          });
+        };
+      };
```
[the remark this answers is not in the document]
YorikSar: Good idea. Will add this.

YorikSar:
@thufschmitt Since your review I have:
* added a higher level wrapper that provides the "flake app" directly, so that boilerplate is gone
* made `nickel-nix-internals` input implicit, but using user's `nixpkgs`, so user doesn't need to know about naked stdenv
* fixed tests: they were using wrong version of nickel-nix input, so moved override to separate call of `nix flake lock`
* added topiary's cache to all flakes to not rebuild it

Please take another look.

thufschmitt (approved):
Thanks, and sorry for the delay in the review

PR #66, opened 2023-05-24
Format all Nickel files with Topiary

YorikSar:


thufschmitt (approved):
Approving and merging to quickly unblock #67, but we'll need to very quickly add a CI check for the formatting in order to keep this property. Maybe also a pre-commit hook if it's cheap enough (which I hope it is).

I'll trust you for only running Topiary, and I'll trust Topiary for not introducing any semantic change because I really don't want to review this line-by-line :stuck_out_tongue:

thufschmitt:
Erk, it didn't re-target the branch when I closed #65 :unamused:

thufschmitt:
Ah, that's because it hadn't deleted the branch. Any way, I've opened (and merged) #70 to fix that

PR #67, opened 2023-05-24
Add Haskell devshell template with stack

YorikSar:
- to use stack's version in the wrapper we expose versions for all derivations in Nickel
- this example uses `stack`, so user can proceed with `stack new --bare simple` to populate it with sample Haskell code


builders.ncl
```diff
@@ -283,4 +283,63 @@ let { NickelDerivation, .. } = import "contracts.ncl" in
         },
       },
     },
+
+  HaskellStackShell =
+    BashShell
+    & {
+      ghcVersion,
```
thufschmitt, line 290:
```suggestion
      ghcVersion, # User-defined. To keep in sync with the one used by stack
```

thufschmitt (changes requested):
Looks good overall. Just a couple of little comments

lib.nix
```diff
@@ -11,9 +11,12 @@ let
         if (value.type or "" == "derivation")
         then {
           "${typeField}" = "nixDerivation";
-          drvPath = value.drvPath;
-          outputName =
-            value.outputName;
+          inherit
+            (value)
+            drvPath
+            version
```
thufschmitt, line 17: Not all the packages have a `version` field, which breaks (in particular) the `c` template: https://github.com/nickel-lang/nickel-nix/actions/runs/5069730717/jobs/9103769608?pr=67

builders.ncl
YorikSar: done

lib.nix
```diff
@@ -22,6 +25,8 @@ let
     then builtins.map exportForNickel value
     else if (type == "lambda")
     then throw "Can’t export a function"
+    else if (type == "path")
```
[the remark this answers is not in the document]
thufschmitt: I've just inlined the necessary bits of code for now to have it working properly. We can refactor it later if needed

thufschmitt (approved):

PR #73, opened 2023-06-02
lib.nix refactoring

YorikSar:
Several small improvements to make it look better.

### DIFF
--- a/lib.nix
+++ b/lib.nix
@@ -189,9 +189,9 @@
         input =
           if inputTakeFrom == "nickel-nix-internals"
           then flakeInputs.${inputTakeFrom}
-          else if inputTakeFrom == "nixpkgs"
-          then flakeInputs.${inputTakeFrom}.legacyPackages.${system}
-          else flakeInputs.${inputTakeFrom}.packages.${system};
+          else
+            flakeInputs.${inputTakeFrom}.legacyPackages.${system}
+            or flakeInputs.${inputTakeFrom}.packages.${system};
 
         pkgPath = declaredInputs.${inputId}.path or [inputId];
       in
--- a/lib.nix
+++ b/lib.nix
@@ -8,10 +8,7 @@
   # Export a Nix value to be consumed by Nickel
   typeField = "$__nixel_type";
 
-  isInStore = path: let
-    sd = builtins.storeDir;
-  in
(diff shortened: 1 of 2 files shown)


YorikSar:
Looks like adding `-c true` to `nix develop` works! :)

lib.nix
```diff
@@ -192,9 +189,9 @@ let
         input =
           if inputTakeFrom == "nickel-nix-internals"
           then flakeInputs.${inputTakeFrom}
-          else if inputTakeFrom == "nixpkgs"
-          then flakeInputs.${inputTakeFrom}.legacyPackages.${system}
-          else flakeInputs.${inputTakeFrom}.packages.${system};
+          else
+            flakeInputs.${inputTakeFrom}.legacyPackages.${system}
+            or flakeInputs.${inputTakeFrom}.packages.${system};
```
yannham, line 194: I think this can, in theory, have a different semantics than the original code if the input is different from `nixpkgs` but defines both `legacyPackages` and `packages`. In that case, the old code would select the one from `packages`, will the new code will select the one from `legacyPackages`. However, in practice, I have no idea if this situation even makes sense: can it happen that another flake than `nixpkgs` defines a `legacyPackages` output? Even in this case, I'm not sure the previous behavior was the right one either. I just wanted to point this out

yannham (approved):

lib.nix
YorikSar: Any flake can in theory define both `packages` and `legacyPackages`, but in practice I'm yet to see non-`nixpkgs` flake with `legacyPackages`, because the difference between is quite obscure. That said, hardcoding `nixpkgs` seems worse because if such flake is used somewhere, there would be no way of using it here.

YorikSar:
@yannham I've just rebased the PR on top of master with no changes. GitHub said that it needs a rebase to be merged. I have commit rights, but I try to wait for reviews before merging. Maybe I should've landed this one after your approval and a couple of quiet days.

thufschmitt:
@YorikSar I think it's reasonable for you to merge once you got an approval. Let's keep this process light :)

yannham:
Right, I'll try to just comment - and not approve - on PR that I feel less confident to endorse (if there's Nix dark magic involved, for example), so that we can keep the 1 approval is sufficient.

PR #74, opened 2023-06-02
Move tests to the flake

YorikSar:
Create "devshell" for each test that can be executed with `nix develop` and call it from GitHub workflows.
This allows to run tests locally with high confidence that they will pass on CI.

Depends on #73

### DIFF
--- a/.github/workflows/examples.yml
+++ b/.github/workflows/examples.yml
@@ -45,18 +45,14 @@ jobs:
     steps:
     - uses: actions/checkout@v3
       with:
-        path: nickel-nix-repo
         fetch-depth: 0
 
     - name: Setup
-      uses: ./nickel-nix-repo/.github/actions/common-setup
+      uses: ./.github/actions/common-setup
       with:
         SECRET_GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         CACHIX_TOKEN: ${{ secrets.CACHIX_TOKEN }}
 
     - name: Build
       run: |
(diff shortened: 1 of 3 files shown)


YorikSar:
@yannham @thufschmitt I'd appreciate a review on this one as well.

thufschmitt (changes requested):
I like the idea of moving that out of the workflows file, but making these tests devshells feels a bit weird. We don't care about the shell part at all, all we care about is running the test script. Could we instead export these scripts as `packages` so that we can do `nix run .#tests.templates.c` or something like that?

Also, we could have only one script that takes the template name as argument and run it through `nix run .#test.templates c` (though I don't have a strong opinion here)

YorikSar:
I got three idea of using devshells for this from home-manager: https://github.com/nix-community/home-manager/blob/master/flake.nix#L90 - for some reason they didn't do them as packages or apps. I see how strange this decision looks without context now.
I'll move them to packages (or apps? nobody  would use them as packages, right?).

I like having this enumeration of tests in Nix, it leaves less logic to differ between CI and dev environment. But we could have both approaches: attributes for CI and argument in case it's more comfortable for dev.

YorikSar:
Oh, right. Having them as shells allows to easily run them without `nix run` command using `nix-shell`. That's why they went this way most likely. It doesn't apply to us at all.

YorikSar:
@thufschmitt I've moved tests to apps, now it's clear that you can only run them, but not build them.

thufschmitt (approved):
Thanks!

PR #77, opened 2023-06-07
flake: Update inputs

YorikSar:
Pull in new Nickel that includes newer version of Topiary so that it can be pulled from the cache instead of rebuilt on each CI call.


thufschmitt (approved):

YorikSar:
Note that Topiary gets rebuild on each build anyway, I'm figuring out the reason with their team.

PR #78, opened 2023-06-08
Add nickel-nix cachix substituter

YorikSar:
We'll keep own copy of all dependencies in it. Should help with not rebuilding them in CI.

### DIFF
--- a/.github/actions/common-setup/action.yml
+++ b/.github/actions/common-setup/action.yml
@@ -3,12 +3,21 @@ inputs:
   SECRET_GITHUB_TOKEN:
     required: true
     description: 'Github Secret Token'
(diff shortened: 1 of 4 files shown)


YorikSar:
@thufschmitt Looks like this cut down CI time from ~10 min for each build to ~3-4 min. It feels like it's dominated by eval and download now.

yannham (approved):
So,you went with a non-Tweag cachix instance in the end :slightly_smiling_face: might be the right move to avoid filling up Tweag's one indeed.

YorikSar:
nickel-lang org in Cachix has been switched to "Open Source" plan now, with 5 GiB limit.

PR #79, opened 2023-06-14
Don’t require --impure

### LINKED ISSUE #15: Remove the need for `--impure`
The current version of nickel-nix requires `--impure` to run, because Nix loses some information when passing derivations to the Nickel world, and then back from the Nickel world.

@thufschmitt started an attempt (https://github.com/thufschmitt/nickel-nix/tree/nicke-shell-pure), which works, but is ugly. We can work our way from here to a cleaner solution.

YorikSar:
Keep trace of all the derivations encountered during the export so that we don’t have to import them again.

The implementation is really ugly, so not sure whether that’s worth it, but at least it’s possible 💪🏻

Rebased and adapted from https://github.com/thufschmitt/nickel-nix/tree/nicke-shell-pure
Depends on #73 ad #74
Closes #15

### DIFF
--- a/lib.nix
+++ b/lib.nix
@@ -10,6 +10,20 @@
 
   isInStore = lib.hasPrefix builtins.storeDir;
 
+  valueWithContext = {
+    create = value: context: {inherit value context;};
+    emptyContext = value: valueWithContext.create value {};
+    mergeContexts = lib.recursiveUpdateUntil (p: _: _: lib.length p < 2);
+    setAttr = set: name: value: {
+      value = set.value // {${name} = value.value;};
+      context = valueWithContext.mergeContexts set.context value.context;
+    };
+    prependToList = value: list: {
+      value = [value.value] ++ list.value;
+      context = valueWithContext.mergeContexts value.context list.context;
+    };
+  };
+
   exportForNickel = value: let
     type = builtins.typeOf value;
   in
@@ -18,29 +32,47 @@
       (
         if (value.type or "" == "derivation")
         then
-          {
-            "${typeField}" = "nixDerivation";
-            inherit
-              (value)
-              drvPath
-              outputName
-              ;
-            outputPath = value.outPath;
+          valueWithContext.create (
+            {
+              "${typeField}" = "nixDerivation";
+              inherit
+                (value)
+                drvPath
+                outputName
+                ;
+              outputPath = value.outPath;
+            }
+            // (
+              if value ? version
+              then {inherit (value) version;}
+              else {}
+            )
+          ) {
+            ${builtins.unsafeDiscardStringContext value.drvPath}.${value.outputName} = value;
           }
-          // (
-            if value ? version
-            then {inherit (value) version;}
-            else {}
+        else
+          builtins.foldl'
+          (
+            acc: eltName:
+              valueWithContext.setAttr acc eltName (exportForNickel value.${eltName})
           )
-        else builtins.mapAttrs (_: exportForNickel) value
+          (valueWithContext.emptyContext {})
+          (builtins.attrNames value)
       )
     else if (type == "list")
-    then builtins.map exportForNickel value
+    then
+      builtins.foldl'
+      (
+        acc: elt:
+          valueWithContext.prependToList (exportForNickel elt) acc
+      )
+      (valueWithContext.emptyContext [])
+      value
     else if (type == "lambda")
     then throw "Can’t export a function"
     else if (type == "path" && isInStore value)
-    then builtins.toString value
-    else value;
+    then valueWithContext.emptyContext (builtins.toString value)
+    else valueWithContext.emptyContext value;
 
   # Take a symbolic derivation (a datastructure representing a derivation), as
   # produced by Nickel, and transform it into valid arguments to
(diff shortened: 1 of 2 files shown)


lib.nix
```diff
@@ -13,29 +29,47 @@ let
...
+                ;
+              outputPath = value.outPath;
+            }
+            // (
+              if value ? version
+              then {inherit (value) version;}
+              else {}
+            )
+          ) {
+            ${builtins.unsafeDiscardStringContext value.drvPath} = value;
```
thufschmitt, line 48: We need the key here to also include the output name (so that `gcc.out` and `gcc.dev` are represented differently). Maybe represent the context as `{ drvPath = { outputName = { value; }; }` and use `lib.recursiveUpdate` to merge them

yannham:
Maybe I should ask that on the issue instead, but as a very bold and uninformed question maybe: what is the expected benefit of not requiring `--impure`? Or rather, what is the cost - beside having to write `--impure` - currently of not tracking deps like this PR tries to do?

I'm really not saying this isn't useful, I'm genuinely asking, to try to see the cost/benefit balance. FWIW I don't find the added complexity to be awful - it's noticeably worse, but it doesn't make the code unreadable either -.

YorikSar:
@yannham  `--impure` applies to the evaluation as a whole, not just one place where we actually need it. It's better to avoid it unless user really want it, especially since we can avoid it, as this change shows.

thufschmitt:
@yannham it doesn't matter too much indeed, except for two reasons:

1. From a messaging perspective, needing `--impure` is a pretty bad signal that we'd rather avoid. Not too big a deal yet, but now is also the right time to fix it
2. `--impure` disables the evaluation caching, which can have a drastic impact when using things like direnv, especially on a project like nickel-nix which requires IFD.

yannham:
> From a messaging perspective, needing --impure is a pretty bad signal that we'd rather avoid. Not too big a deal yet, but now is also the right time to fix it

Ah, that's what I assumed too, but wasn't sure. It makes sense, messaging is also important.

> --impure disables the evaluation caching, which can have a drastic impact when using things like direnv, especially on a project like nickel-nix which requires IFD.

This a compelling argument as well! Thanks for the explanations.

lib.nix
YorikSar: I've added this, but now I'm not sure if I should have. We're accessing them as `context.${value.drvPath}.${value.outputName or "out"}`, so even if `gcc.dev` overrides `gcc.out`, we'd still access it as `gcc.dev.out` which is supposed to be the same as `gcc.out`.
That said, making this explicit removes such confusion and is probably safer for the future.

lib.nix
```diff
@@ -1,9 +1,25 @@
...
+  valueWithContext = value: context: {inherit value context;};
+  valueWithNoContext = value: valueWithContext value {};
+  setAttrInSetWithContext = set: name: value: {
+    value = set.value // {${name} = value.value;};
+    context = set.context // value.context;
+  };
+  appendToListWithContext = l: r: {
+    value = [l.value] ++ r.value;
+    context = l.context // r.context;
+  };
```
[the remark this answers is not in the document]
YorikSar: Changed.

lib.nix
```diff
@@ -10,6 +10,20 @@
 
   isInStore = lib.hasPrefix builtins.storeDir;
 
+  valueWithContext = {
+    create = value: context: {inherit value context;};
```
yannham, line 14: Very small nitpick, but I feel like `new` is more common as a constructor name. Very unimportant, though

YorikSar: @yannham Do you still have some review you want to share? If so, could you?

YorikSar:
I've rebased this on top of the master and marked it as ready for review, because all dependencies are merged. Looking forward to the review!

yannham (commented):
So, if I understand correctly, we track by hand all the contexts that are attached to strings that could appear somewhere inside the packages exporter from Nix to Nickel, which we can then restore when we get back some of those package somewhere in the resulting derivation produced by Nickel.

Might be a stupid question, but isn't already able to perform all the tracking part? Could we do something using `builtins.getContext` (and maybe a traversal) instead of tracking this by hand explicitly?

thufschmitt:
@yannham the use of “context” here might be slightly unfortunate here because what we're tracking isn't exactly the contexts (in the Nix sense). What we do is that we change a bit the evaluation model to make the roundtrip less lossy:

- What we do on master is that we just (well, mostly) pass in the drv path to Nickel and then back to Nix, and re-import it.
- What this PR does is that it keeps the original value on the Nix side and reuses that when getting the drv path back from Nickel.

The big advantage of this is that Nix isn't willing to import arbitrary paths in pure mode (and it doesn't know that these drv paths aren't arbitrary but that we've just built them), which is why the current solution requires `--impure`. By keeping track of the original value, we avoid that

thufschmitt:
Otoh, we can abuse a Nix bug where a clever use of `builtins.appendContext` allows to magically bypass this. But I'm not sure whether that's a long-term situation

yannham (approved):
I'm not sure to fully understand yet why we have to do the book-keeping by hand and use `valueContext`. Wouldn't a function `buildContext`, called on the final `exportedPkgs`, be implementable, by looking for derivation and strings (and here use `builtins.getContext`) ? In any case, I might be missing something, and this is possible to do as a second step. On the other hand, I think this is a non-trivial change which impacts the implementation of Nixel, so I would strongly recommend to document it right now, in this PR.

YorikSar:
@yannham I've added a comment describing the structure and the process involved with this structure. Please take check if it's fine.

> Wouldn't a function buildContext, called on the final exportedPkgs, be implementable, by looking for derivation and strings (and here use builtins.getContext) ?

No, it wouldn't. We need the original derivation objects on the Nix side which are lost as soon as we convert them to the format compatible with Nickel (or rather JSON). The `context` here is not the same as "string context" used in Nix and Nickel. It's not attached to values, it's only accumulated during the conversion.

yannham:
Ok, I think I get it now, thanks for the explanations. @YorikSar the comment goes a long way yes, it's good to go on my side :+1:

PR #80, opened 2023-06-20
Replace substituters with nickel-nix.cachix.org

YorikSar:
in all templates and the example.
We push all paths not in cache.nixos.org to nickecl-nix.cachix.org anyway, so no need to keep multiple substituters in flake.nix files.

Don't change nickel-nix flake.nix though, because we don't want to rebuild everything on dependency updates.


thufschmitt (approved):
Nice and concise :)

PR #81, opened 2023-06-20
Add a function to generate flake outputs

YorikSar:
Reduces flake.nix boilerplate on the user side.

Fixes https://github.com/nickel-lang/nickel-nix/issues/71.


thufschmitt (approved):
That's much nicer, thanks :)

I don't know whether we'll keep this exact interface, but at least it's much easier to evolve now

PR #83, opened 2023-06-28
Add a `packages` field to the shells

thufschmitt:
Allows easily adding stuff to `$PATH`.

Eventually this should hook into `nativeBuildInputs` (or equivalent) once https://github.com/nickel-lang/nickel-nix/issues/72 is fixed, but for now just setting the `PATH` will do.

### DIFF
--- a/contracts.ncl
+++ b/contracts.ncl
@@ -275,7 +275,6 @@ from the Nix world) or a derivation defined in Nickel.
         | default
         = {},
       "%{type_field}" | force = "nickelDerivation",
-      ..
     },
 
   Params | doc "The parameters provided to the Nickel expression"
--- a/contracts.ncl
+++ b/contracts.ncl
@@ -275,6 +275,7 @@ from the Nix world) or a derivation defined in Nickel.
         | default
         = {},
       "%{type_field}" | force = "nickelDerivation",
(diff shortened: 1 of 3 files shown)


contracts.ncl
```diff
@@ -275,6 +275,7 @@ from the Nix world) or a derivation defined in Nickel.
         | default
         = {},
       "%{type_field}" | force = "nickelDerivation",
+      ..
```
YorikSar, line 278: Should we define `packages` here?

thufschmitt: That's a good question. I think not, because `packages` is something defined by `BashShell`, so we shouldn't have to know about it here. Similarely, we'll want other shells to be able to extend what we put in here (say have `RustShell` give you a `rustPackage` version to override the one used), so this thing should be arbitrarily extensible.

That raises an interesting issue as I don't know how we can ergonomically represent a closed, but extensible record contract.

thufschmitt: Actually, def4628 makes this Okay-ish for the moment, so we can stick with it I guess

YorikSar: Agree, it looks much better :)

YorikSar (approved):

PR #87, opened 2023-07-05
Radically cleanup the templates

### LINKED ISSUE #86: Get rid of the bazillon templates
We currently expose 9 templates, which are essentially copy-paste of each other (modulo the shell it merges with). We sould just expose one and let users change the shell accordingly

thufschmitt:
Remove all the templates, except for one.
This makes the repository quite simpler, and emphasises the fact that we want something that just works™

We'll certainly want to add some funkier tests, but in a more principled way, and that doesn't leak into the end-user interface.

~Also removes the example for the same reason.~

Fix #86


thufschmitt:
(The merging is blocked since it's expecting the `Test devshell haskell-devshell` job to succeed. But I'll change that to `Test devshells` if we agree on that PR)

test-template.sh
```diff
@@ -0,0 +1,38 @@
...
+  function clean() {
+    rm -rf "${WORKDIR}"
+  }
+  trap clean EXIT
+  pushd "${WORKDIR}"
+
+  nix flake new --template "path:$PROJECT_ROOT" example --accept-flake-config
+
+  pushd ./example
+  sed -i "s/Nickel/$target/" dev-shell.ncl
```
YorikSar, line 18:
```suggestion
  sed -i "s/BashShell/$target/" dev-shell.ncl
```

flake.nix
```diff
@@ -182,62 +166,6 @@
           pkgs.lib.listToAttrs
         ];
 
-        apps.tests = let
```
YorikSar, line 185: Please wrap the test script into app (or many). It's much easier to run `nix run .#tests.all` than to keep in memory the full `nix develop --command bash test-template.sh`.

.github/workflows/templates.yml
```diff
@@ -53,6 +23,6 @@ jobs:
         SECRET_GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         CACHIX_TOKEN: ${{ secrets.CACHIX_TOKEN }}
 
-    - name: Test devshell ${{ matrix.template }}
+    - name: Test devshells
       run: |
-        nix run .#tests.templates.${{ matrix.template }}
+        nix develop --accept-flake-config --command bash test-template.sh
```
YorikSar, line 28: I wonder how long it will take without parallelisation...

YorikSar (changes requested):

.github/workflows/templates.yml
thufschmitt: image

YorikSar: That's running BashShell 14 times ;)

templates/devshells/haskell/dev-shell.ncl
```diff
@@ -1,9 +0,0 @@
-let builders = (import "./nickel.lock.ncl").nickel-nix.builders in
-
-{
-  ghcVersion = "927",
```
YorikSar, line 4: Looks like not all templates are the same after all. Without this line we get this in CI:

```
building '/nix/store/dy3xs2il3qy9f6g92sipk5b9viy1ym4b-nickel-inputs.json.drv'...
nickel-inputs.json> error: missing definition for `ghcVersion`
nickel-inputs.json>     ┌─ /nix/store/3g0ymzm2z6dvls62y0s73g12rb8j63ip-5gm6jqx5lkzfxpslkjmc22b037g48b8i-source/builders.ncl:274:7
nickel-inputs.json>     │
nickel-inputs.json> 274 │       ghcVersion, # User-defined. To keep in sync with the one used by stack
nickel-inputs.json>     │       ^^^^^^^^^^ required here
nickel-inputs.json>     │
nickel-inputs.json>     ┌─ /nix/store/nm7k7izblms5h2rk62dswl420crlpp5g-h0vdahwbi2fbhaccnwik204z8vcmv7fz-source/dev-shell.ncl:3:1
nickel-inputs.json>     │
nickel-inputs.json>   3 │ ╭ {
nickel-inputs.json>   4 │ │   output = {
nickel-inputs.json>   5 │ │     name = "nickel-shell",
nickel-inputs.json>   6 │ │   }
nickel-inputs.json>   7 │ │ }
nickel-inputs.json>   8 │ │ & nixel.builders.HaskellStackShell
nickel-inputs.json>     │ ╰──────────────────────────────────' in this record
```

test-template.sh
```diff
@@ -0,0 +1,38 @@
...
+  nix develop --accept-flake-config --print-build-logs < /dev/null
+  popd
+  popd
+  clean
+}
+
+if [[ -n ${1+x} ]]; then
+  test_one "$1"
+else
+  all_targets=$(nickel export <<<'std.record.fields (import "builders.ncl") |> std.string.join " "')
```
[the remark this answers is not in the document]
thufschmitt: Bleh, thanks. Fixed

examples/c-hello-world/README.md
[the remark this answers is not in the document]
thufschmitt: It's not really an end-to-end example, it's an example of something that we don't care about too much right now (building packages). So I don't think we need to keep it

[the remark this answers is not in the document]
YorikSar: Ok. I think we should still keep in mind that we should bring something like this back in the future.

.github/workflows/templates.yml
thufschmitt: It's indeed a bit slower now :)

image

Do you think that justifies restoring the parallelism?

YorikSar: After that Cachix pushed 116 paths. Maybe that's just cold cache? I've retriggered it here: https://github.com/nickel-lang/nickel-nix/actions/runs/5465840641/jobs/9950269605

thufschmitt: (Or should we just parallelize within a single runner? That might be even faster)

YorikSar: I wonder how hard it would be. Just running multiple Bash jobs would make it impossible to read logs afterwards.
Maybe enabling recursive-nix is not that hard?.. :)

YorikSar: By the way, it looks like CI is not strictly repeatable: we're using latest `nixpkgs` input on each run, so cache is probably getting invalidated more often than it should be. Maybe we should add `--inputs-from ${./.}` to `nix flake lock` there?

thufschmitt: > we're using latest nixpkgs input on each run, so cache is probably getting invalidated more often than it should be

Duh, indeed :grimacing:

thufschmitt:
@YorikSar I've added the example back. PTAL

flake.nix
```diff
@@ -174,60 +158,19 @@
...
-          );
+        apps.run-test =
+          let testScript = pkgs.writeShellApplication {
+            name = "test-templates";
+            runtimeInputs = [ inputs.nickel.packages."${system}".nickel-lang-cli ];
+            text = ''
+              set -euo pipefail
+              cd ${./.}
+              exec ${pkgs.runtimeShell} ./run-test.sh "$@"
+            '';
```
YorikSar, line 169:
```suggestion
            text = builtins.readFile ./run-test.sh;
```

Since shell script requires this `cd` line anyway, it should happen in the script. This would also run additional checks against the script itself (like ShellCheck) so that we can be more confident in it.

YorikSar (changes requested):

run-test.sh
```diff
@@ -0,0 +1,83 @@
...
+      test_one_template $target
+    done
+  fi
+}
+
+test_example () (
+  examplePath=$(realpath "$1")
+  pushd_temp
+  cp -r "$examplePath" .
+  pushd *
```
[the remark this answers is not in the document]
thufschmitt: It will always have one element since it's a fresh dir on which we copy one thing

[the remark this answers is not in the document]
thufschmitt: I've changed it so that we know the name of the directory we `pushd` into. Makes it a bit less scary

flake.nix
thufschmitt: I couldn't find a nice way of `cd`-ing directly from the script (as we can't really know the path of the flakeroot from it). But I've just dropped it as it's fair to just assume that we're already in the right directory

YorikSar (approved):

PR #88, opened 2023-07-06
PoC: Provide inputs as Nickel records

YorikSar:
Generate Nickel record with all packages from each input and allow Nickel code to access them directly without declaring them in `input_spec`.


YorikSar:
I've finally got this to somewhat work. Current limitations:
* have to limit `nixpkgs` input to only supported pkgs because of https://github.com/tweag/nickel/issues/1428 and hack Haskell subset with `recurseForDerivations` fields. It seems we'd either have to evaluate *everything* in `nixpkgs` or somehow pick paths that are the most interesting for us.
* inputs are mixed into `inputs` field, it should be quite straightforward to add them to the lockfile, but then we'll need to pass it into `builders.ncl` somehow.
* files should be generated in separate derivations so that we can cache them and not re-evaluate then every time. I've tried just running `nix-instantiate --eval --offline --extra-experimental-features flakes --store ./store "''${nixSourcePath}" > $out` with `nixSource` set to:
  ```nix
        let
          nickel-nix = (builtins.getFlake "${./.}");
          input = builtins.getFlake "${input}";
        in
          nickel-nix.lib.${system}.collectPkgs input
  ```
  but that yields sandbox related errors. Probably it won't be a huge deal if we generate these files when generating the lockfile.

thufschmitt:
I would wait for 0.1 to be out as there's already a bit of work do be done first (besides the fact that we can't use the latest Nickel because of https://github.com/tweag/nickel/issues/1664). But that would be great if it allows us to improve the Nix interop!

yannham:
@YorikSar @thufschmitt what is the status of this PR? Is this work still relevant, but needs to be rebased, or can we just close it ?

PR #89, opened 2023-07-07
Use nixpkgs stdenv

balsoft:
Use nixpkgs stdenv for setting up environment variables.

Fixes https://github.com/nickel-lang/nickel-nix/issues/72


builders.ncl
```diff
@@ -34,44 +41,29 @@ let { NickelDerivation, Derivation, .. } = import "contracts.ncl" in
...
-    inputs,
-
-    output =
-      {
-        version | default = "0.1",
-
-        # This is required otherwise nix develop will fail with a message that it
-        # only supports bash.
+  NixpkgsPkg
+    | doc m%%%"
```
yannham, line 45: It's customary to use the simplest delimiter for multiline strings, `m%"`. Longer delimiters are mostly used when you have other Nickel mutliline strings or escaped interpolation sequences inside (e.g. when documenting a stdlib function, and you want to include a Nickel example which itself has a mutliline string) and don't want to escape stuff too much. But here, it's just text

```suggestion
    | doc m%"
```

builders.ncl
```diff
@@ -81,195 +73,158 @@ let { NickelDerivation, Derivation, .. } = import "contracts.ncl" in
...
-                |> std.array.fold_left (fun acc value => nix-s%"%{acc}:%{value}"%) first
+                # Intersperse ' ' between environment variable fragments
+                concatStringsSep " " (std.record.values xs)
             ),
-      } | (NickelPkg & { packages | { _ : Derivation } }),
+      } | NickelPkg,
+    },
+
+  Shell
+    | doc m%%%"
```
yannham, line 83:
```suggestion
    | doc m%"
```

builders.ncl
```diff
@@ -34,44 +41,29 @@ let { NickelDerivation, Derivation, .. } = import "contracts.ncl" in
...
-      {
-        version | default = "0.1",
-
-        # This is required otherwise nix develop will fail with a message that it
-        # only supports bash.
+  NixpkgsPkg
+    | doc m%%%"
+      Makes a derivation that runs all the build phases from nixpkgs' stdenv as the `build_command`.
+      Can be controlled with environment variables in the same way as `stdenv.mkDerivation`.
+    "%%%
```
yannham, line 48:
```suggestion
    "%
```

builders.ncl
```diff
@@ -81,195 +73,158 @@ let { NickelDerivation, Derivation, .. } = import "contracts.ncl" in
...
             ),
-      } | (NickelPkg & { packages | { _ : Derivation } }),
+      } | NickelPkg,
+    },
+
+  Shell
+    | doc m%%%"
+      A derivation that is to be used as a shell, e.g. with `nix develop`.
+      Analogous to `mkShell`.
+    "%%%
```
yannham, line 86:
```suggestion
    "%
```

builders.ncl
```diff
@@ -1,5 +1,12 @@
-let { NickelDerivation, Derivation, .. } = import "contracts.ncl" in
-
+let { NickelDerivation, Derivation, NixString, .. } = import "contracts.ncl" in
+
+let concatStringsSep = fun sep values =>
```
yannham, line 3: In general we tend to use snake_case for identifiers, and CamelCase for types/contracts.

```suggestion
let concat_strings_sep = fun sep values =>
```

builders.ncl
```diff
@@ -81,195 +73,158 @@ let { NickelDerivation, Derivation, .. } = import "contracts.ncl" in
...
+      output = {
+        name | default = "shell",
+        version | default = "dev",
+
+        env.buildCommand = nix-s%"
+          echo "This derivation is not supposed to be built" 1>&2 1>/dev/null
+          exit 1
+        "%,
+        env.shellHook = concatStringsSep "\n" (std.record.values hooks),
+        structured_env.buildInputs = packages,
```
yannham, line 100: Same thing about casing. Unless those are values that are then exported to Nix? Even there, I wonder if we shouldn't do the translation by hand between snake_case and camelCase. Otherwise it's going to be really confused as to which casing to use. Related discussion: https://github.com/tweag/nickel/discussions/1311

balsoft: This is passed to nixpkgs stdenv: https://nixos.org/manual/nixpkgs/stable/#var-stdenv-buildInputs .

yannham: I see, we'll have to think about this at some point, but this PR might not be the right scope. If it's passed onto Nixpkgs, we don't have much choice, at least for now.

builders.ncl
```diff
@@ -1,5 +1,12 @@
-let { NickelDerivation, Derivation, .. } = import "contracts.ncl" in
-
+let { NickelDerivation, Derivation, NixString, .. } = import "contracts.ncl" in
+
+let concat_strings_sep = fun sep values =>
+  if std.array.length values == 0 then "" else
+    let first = std.array.first values in
+    values
+    |> std.array.drop_first
+    |> std.array.fold_left (fun acc value => nix-s%"%{acc}%{sep}%{value}"%) first
```
yannham, line 8: I missed that on the first review, but you can use `std.array.reduce_left` for folding starting with the first element, without having to `drop_first` manually

yannham (approved):
LGTM, as far as I can tell on the Nickel side :+1:  it's nice to see that provided shells also seem less boilerplaty.

builders.ncl
yannham: https://nickel-lang.org/stdlib/std-array/#reduce_left

balsoft: Thanks, done!

yannham:
@balsoft Should I just merge this, or do I let you push the green button? Or do you prefer a second review from @thufschmitt ?

balsoft:
I would love a second review, as it's quite a big change.

builders.ncl
```diff
@@ -81,195 +70,158 @@ let { NickelDerivation, Derivation, .. } = import "contracts.ncl" in
...
             (
               fun _n xs =>
-                # Intersperse ':' between environment variable fragments
-                let values = std.record.values xs in
-                let first = std.array.first values in
-                values
-                |> std.array.drop_first
-                |> std.array.fold_left (fun acc value => nix-s%"%{acc}:%{value}"%) first
+                # Intersperse ' ' between environment variable fragments
+                concat_strings_sep " " (std.record.values xs)
```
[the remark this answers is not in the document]
thufschmitt: Actually, using `structuredAttrs` mean that we can directly pass a record as `buildInputs` (it'll be passed as an associative array rather than a numerically-indexed array, but that doesn't change anything). So we can get rid of the concat dance altogether

builders.ncl
```diff
@@ -81,195 +70,158 @@ let { NickelDerivation, Derivation, .. } = import "contracts.ncl" in
...
+        version | default = "dev",
+
+        env.buildCommand = nix-s%"
+          echo "This derivation is not supposed to be built" 1>&2 1>/dev/null
+          exit 1
+        "%,
+        env.shellHook = concat_strings_sep "\n" (std.record.values hooks),
+        structured_env.buildInputs = packages,
+      } | NickelPkg,
+    },
```
thufschmitt, line 99: There's a bit of confusion here between what's an input and what's an output. But it's already the case atm, so it's more a note for later than a real concern with this PR

thufschmitt (changes requested):
I like that change quite a lot. I'd like to try using `__structuredAttrs` before merging as it feels like it could make the implementation cleaner though

builders.ncl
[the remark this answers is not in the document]
balsoft: > Could we instead use structuredAttrs? stdenv is supposed to be able to handle that just fine now, and that would be nicer than doing this kind of concatenation ourselves

It gets annoying fast. Since `nix-s%"..."%` gets translated to a record (`{ prefix = ..., tag = ..., fragments = ... }`) internally, we have to distinguish between that sort of record and actual records of nix strings. And when I did that, I got `error: boost::bad_format_string: format-string is ill-formed` from Nix :D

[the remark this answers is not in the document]
thufschmitt: OK, let's keep away from it for now then.

> I got error: boost::bad_format_string: format-string is ill-formed from Nix :D

:partying_face:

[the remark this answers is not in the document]
balsoft: Actually, figured it out. I've kept `env` and `structured_env`.

thufschmitt:
#96 fixed the “use nixpkgs stdenv” part, so the remaining bit here is the `WithInput` shells (which automatically add their inputs to `buildInputs`), and these will be superseded by #100. So closing this. @balsoft feel free to reopen if there's something more to pick from this PR that I've missed

PR #98, opened 2023-07-27
Reorganize the lib files

thufschmitt:
Move the library files to a subfolder and use a single entry point for the whole of Nixel.

Makes it quite easier to understand what's going on

Depends on #97


flake.nix
```diff
@@ -65,7 +65,7 @@
       flake.outputsFromNickel = path: inputs: {
         systems ? flake-utils.lib.defaultSystems,
         lockFileContents ? {
-          nickel-nix = self.lib.x86_64-linux.lockFileContents;
+          nickel-nix = "${./lib}/nix.ncl";
```
YorikSar, line 68:
```suggestion
          nickel-nix = "${./.}/lib/nix.ncl";
```

Otherwise `${./lib}` makes a separate copy of `lib` in the store that is less likely to be present - it will be copied only when this value is evaluated, not as part of flake dependency resolution.

YorikSar (changes requested):

YorikSar (approved):

PR #101, opened 2023-07-28
Simplify the schemas used

thufschmitt:
Rework the schemas to make the logic clearer, remove a lot of cruft and document things a bit more

Depends on #100

### DIFF
--- a/lib/contracts.ncl
+++ b/lib/contracts.ncl
@@ -45,6 +45,9 @@ in
   # instead?
   System = { arch | String, os | String },
 
+  NullOr = fun c label value =>
+    if value == null then value else std.contract.apply c label value,
+
   # TODO: more precise contract
   Derivation
     | doc m%"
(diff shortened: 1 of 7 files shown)


lib/contracts.ncl
```diff
@@ -45,6 +44,11 @@ in
   # instead?
   System = { arch | String, os | String },
 
+  NullOr
```
yannham, line 47: Ok, we REALLY need this to be a builtin type. You shouldn't have to define that by yourself in 1.0.0+

lib/contracts.ncl
```diff
@@ -45,6 +44,11 @@ in
...
+    | doc "Make a contract nullable"
+    = fun c label value =>
```
yannham, line 49: I like long names :p 

```suggestion
    = fun contract label value =>
```

lib/contracts.ncl
```diff
@@ -154,193 +158,119 @@ from the Nix world) or a derivation defined in Nickel.
         let { fragments, .. } = std.contract.apply NixSymbolicString label value in
         mk_nix_string fragments,
 
-  NickelDerivation
+  NixDerivation
     | doc m%"
         The basic, low-level interface for a symbolic derivation. A
-        NickelDerivation is intenteded to be passed (exported) to the Nix side,
+        NixDerivations is intenteded to be passed (exported) to the Nix side,
```
yannham, line 164:
```suggestion
        NixDerivation is intended to be passed (exported) to the Nix side,
```

yannham (commented):
It's just my two cents, but I would stick to Nickel naming convention (snake_case for identifiers), except those are attributes that are going to go untransformed on the Nix side, which then needs to respect some schema, morally. Also, I would avoid dash such as in `nickel-nix` (dashes are allowed more like an escape hatch because external tools might enforce them within names, but internally I personally think they should be avoided, as we already have snake_case's `_`). Otherwise it's becoming a bit hard to follow: Nickel stdlib's functions (or maybe, later, external libraries) use snake_case, then Nixel symbols use camelCase, etc.

thufschmitt:
> It's just my two cents, but I would stick to Nickel naming convention

Ooh, yes, thanks for reminding me of that :) That should be fixed (at least for the stuff changed in that PR, there might be some older inconsistencies here and there).

> I would avoid dash such as in nickel-nix (dashes are allowed more like an escape hatch because external tools might enforce them within names, but internally I personally think they should be avoided, as we already have snake_case's _)

Fair enough. I'll wait for us to have a definite name before changing things

PR #106, opened 2023-07-28
Split the shells into a `build` and `dev` variant

thufschmitt:
Rather than exposing a single shell under `output`, split them between
- `dev` (the default one)
- `build` (a trimmed-down version of `dev` for contexts where we just want to build things and don't need all the dev helpers).

 Depends on #101

### DIFF
--- a/lib/builders.ncl
+++ b/lib/builders.ncl
@@ -45,7 +45,6 @@ in
       Can be controlled with environment variables in the same way as `stdenv.mkDerivation`.
     "%
     = {
-      output = {
         name,
         version,
         build_command = {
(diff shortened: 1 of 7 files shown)


lib/builders.ncl
```diff
@@ -80,20 +77,24 @@ in
...
-  BashShell = Shell & {
-    output.packages = {
-      bash = lib.import_nix "nixpkgs#bash",
+  BashShell = {
+    build = Shell & {
+      packages = {
+        bash = lib.import_nix "nixpkgs#bash",
+      },
     },
+    dev = Shell,
```
aspiwack, line 87: Not
```suggestion
    dev = build,
```
?

aspiwack (commented):
I like how little the `builders.ncl` file must change.

lib/builders.ncl
thufschmitt: Oh, indeed

PR #107, opened 2023-07-28
Run the template tests in parallel

thufschmitt:
Make the tests faster by not running them sequentially. Makes the test suite wall time drop from 2m30 to 30s on my machine.


thufschmitt:
Unsurprisingly, the time gain is smaller on the CI (20mins -> 7mins30 roughly). But still appreciable :)

PR #117, opened 2023-09-04
Format Nix files with Alejandra

YorikSar:
Also add a check to the flake and run it on CI to keep format clean
in the future


flake.nix
```diff
@@ -151,6 +152,11 @@
             pkgs.parallel
```
thufschmitt, line 152:
```suggestion
            pkgs.parallel
            pkgs.alejandra
```

thufschmitt (approved):
Just needs adding it to the dev shell, but lgtm otherwise

PR #118, opened 2023-09-04
Autogenerate nickel.lock.ncl

### LINKED ISSUE #105: Check that the Nickel lockfile is in sync with the flake
Nixel uses a `nickel-lock.ncl` file which is manually generated (by running `nix run .#regenerate-lockfile`) and has to be kept in sync with the flake inputs.
Failing to do so leads to a confusing errors since the paths mentioned in the lockfile won't be in the Nix sandbox, so Nickel will fail with a "file not found" error (although they do exist in the filesystem).

To make this more understandable, we could check during the evaluation that the lockfile is up-to-date, and provide a more sensible error message if not.

### LINKED ISSUE #116: Generate the Nickel lockfile on the fly on the sandbox
Organist relies on a Nickel lockfile (`nickel.lock.ncl`) that has to be kept in sync with the flake inputs (through running `nix run .#regenerate-lockfile`). That's slightly suboptimal in terms of ergonomy, but there's not too much we can do for now.

The part where it gets annoying though, is that it causes a bootstrap problem: We'd like the entirety of the flake output to be defined in Nickel, but we can't because `apps.${system}.regenerate-lockfile` must still evaluate properly even when the lockfile is outdated or ill-formed (and thus the Nickel expression doesn't evaluate).
Since the generation of the Nickel lockfile is a pure function of the flake inputs, we could however do it systematically in the sandbox before calling Nickel. That would allow us to expose `regenerate-lockfile` from the Nickel expression in `project.ncl` like the rest and simplify the flakes output handling in the code.

For the LSP, we still need a static `nickel.lock.ncl` of course, so we'll want the builder to warn if it is outdated, and prompt the user to run `nix run .#regenerate-lockfile` in that case (or that could even be done automatically as part of the shell hook).

YorikSar:
Re-generate expected contents of nickel.lock.ncl on every importNcl call, check if it matches one in the sources and generate a new source tree with new nickel.lock.ncl if it doesn't. Note that expected lockfile contents depend only on flake parameters. Also note that when lockfile contents don't match the expected value, we'll make a full copy of the source tree during evalutaion of Nickel file.

Fixes #105
Fixes #116

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -28,9 +28,7 @@
         welcomeText = ''
           You have created a devshell that is built using nickel!
 
-          First run `nix run .#regenerate-lockfile` to fill `nickel.lock.ncl` with proper references.
-
-          Then run `nix develop` to enter the dev shell.
+          You can run `nix develop` to enter the dev shell.
         '';
       };
 
@@ -55,7 +53,7 @@
             apps.regenerate-lockfile = lib.regenerateLockFileApp lockFileContents;
           }
           // pkgs.lib.optionalAttrs (builtins.readDir path ? "project.ncl") rec {
-            nickelOutputs = lib.importNcl path "project.ncl" inputs;
+            nickelOutputs = lib.importNcl path "project.ncl" inputs lockFileContents;
             packages.default = nickelOutputs.packages.default or {};
             devShells = nickelOutputs.shells or {};
           });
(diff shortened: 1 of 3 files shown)


YorikSar:
@thufschmitt Now we generate and compare lockfile contents in Nix and don't use additional IFD for that.
Note that lockfile contents currently contain an additional newline in the end. I'll fix it a bit later.

thufschmitt (approved):
Awesome, thanks @YorikSar 

The performance hit isn't there anymore indeed. Thanks!

PR #128, opened 2023-09-05
Refresh flake.lock

YorikSar:
https://github.com/nickel-lang/organist/pull/122 landed without relevant flake.lock changes. Land these changes and add a check to prevent this from happening in the future.

### DIFF
--- a/.github/workflows/templates.yml
+++ b/.github/workflows/templates.yml
@@ -28,6 +28,14 @@ jobs:
       run: |
         nix flake check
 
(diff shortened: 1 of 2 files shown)


YorikSar:
Here's an example of the check failing without flake.lock change: https://github.com/nickel-lang/organist/actions/runs/6088127317/job/16518145287

thufschmitt (approved):
Bleh, thanks

PR #133, opened 2023-09-08
Add upstream Nickel cache to CI

YorikSar:
While #132 removed it from the flake because flake users should be able to fetch everything from organist cache, CI still would benefit from fetching from upstream cache to avoid unnecessary rebuilds when updating Nickel, and then push these paths to the organist cache.


thufschmitt (approved):
Oooh indeed, thanks for catching that.

Maybe we should fix that upstream (not have `nix init` pick the substituters)

PR #134, opened 2023-09-09
move buildLockFile and buildLockFileContents to lib.nix

BurNiinTRee:
previously `buildLockFile` and
`buildLockFileContents` were only exposed through
the flake, even though `lib/lib.nix` depends
on them. This made using organist without flakes
more complicated, as those two function definitions had to be copied in the consuming nix code.

Now using organist without flakes is as simple as

```nix
let
  pkgs = import <nixpkgs> {};
  organist = builtins.fetchTarball {
    ...
  };
  organistLib = pkgs.callPackage (organist + "/lib/lib.nix") {
    flakeRoot = organist;
    inherit organistLib;
  };
  lockFileContents = {
    organist = "${organist}/lib/nix.ncl";
  };
in {
  organist = organistLib.importNcl
    ./.
    "project.ncl"
    {nixpkgs = pkgs;}
    lockFileContents;
  regenerate-lockfile = organistLib.buildLockFile lockFileContents;
}
```

Should `lockFileContents` also be moved there?
Is there a reason a downstream user would set it
to something else?

This also removes the need to pass organistLib
to itself.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -68,48 +68,6 @@
       in {
         lib.importNcl = lib.importNcl;
 
-        # Helper function that generates contents for "nickel.lock.ncl", see buildLockFile.
-        lib.buildLockFileContents = contents: let
-          lib = pkgs.lib;
-          getLinesOne = indent: name: thing:
-            if lib.isAttrs thing
-            then
-              [
-                (indent + (lib.optionalString (name != null) "${name} = ") + "{")
-              ]
-              ++ lib.mapAttrsToList (getLinesOne (indent + "  ")) thing
-              ++ [
-                (indent + "}" + (lib.optionalString (name != null) ","))
-              ]
-            else [''${indent}${name} = import "${builtins.toString thing}",''];
-        in
-          lib.concatLines (lib.flatten (getLinesOne "" null contents));
-
-        # A script that generates contents of "nickel.lock.ncl" file from a recursive attribute set of strings.
-        # Example inputs:
-        #   {
-        #     organist = {
-        #       builders = "/nix/store/...-source/builders.ncl";
-        #       contracts = "/nix/store/...-source/contracts.ncl";
-        #       nix = "/nix/store/...-source/nix.ncl";
-        #     };
-        #   }
-        # Result:
-        #   {
-        #     organist = {
-        #       builders = import "/nix/store/...-source/builders.ncl",
-        #       contracts = import "/nix/store/...-source/contracts.ncl",
-        #       nix = import "/nix/store/...-source/nix.ncl",
-        #     },
-        #   }
-        lib.buildLockFile = contents:
-          pkgs.writeShellApplication {
-            name = "regenerate-lockfile";
-            text = ''
-              cat > nickel.lock.ncl <${builtins.toFile "nickel.lock.ncl" (self.lib.${system}.buildLockFileContents contents)}
-            '';
-          };
-
         # Flake app to generate nickel.lock.ncl file. Example usage:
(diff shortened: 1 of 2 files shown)


lib/lib.nix
```diff
@@ -84,7 +125,7 @@
     };
 
     lockfilePath = "${sources}/nickel.lock.ncl";
-    expectedLockfileContents = organistLib.buildLockFileContents lockFileContents;
+    expectedLockfileContents = buildLockFileContents lockFileContents;
```
thufschmitt, line 128: Since we don't have this dependency on `organistLib` any more, can you drop it from the arguments of this file altogether? That'll make the whole thing a bit simpler

thufschmitt (changes requested):
Nice, thanks :)

Just one small suggestion, and it's good to go

lib/lib.nix
```diff
@@ -139,4 +179,4 @@
     {rawNickel = nickelResult;}
     // (importFromNickel flakeInputs system baseDir (builtins.fromJSON
         (builtins.unsafeDiscardStringContext (builtins.readFile nickelResult))));
-in {inherit importNcl;}
+in {inherit importNcl buildLockFile;}
```
YorikSar, line 182: Please put it on separate lines and also export `buildLockFileContents`.

```suggestion
in {
  inherit
    importNcl
    buildLockFile
    buildLockFileContents
    ;
}
```

flake.nix
```diff
@@ -59,57 +59,13 @@
     // flake-utils.lib.eachDefaultSystem (
       system: let
         lib = pkgs.callPackage ./lib/lib.nix {
-          inherit system;
```
YorikSar, line 62: This argument is still present and in use, please keep it around.

flake.nix
```diff
@@ -59,57 +59,13 @@
...
-        #   }
-        lib.buildLockFile = contents:
-          pkgs.writeShellApplication {
-            name = "regenerate-lockfile";
-            text = ''
-              cat > nickel.lock.ncl <${builtins.toFile "nickel.lock.ncl" (self.lib.${system}.buildLockFileContents contents)}
-            '';
-          };
-
         # Flake app to generate nickel.lock.ncl file. Example usage:
```
YorikSar, line 69: I think we shouldn't stop there and move `regenerateLockFileApp` also to `lib/lib.nix` and simplify this flake file by just calling `lib = pkgs.callPackage lib/lib.nix {...}`. Then also pieces that were already moved to `lib/lib.nix` will still be accessible via flake.

YorikSar (changes requested):
Great, I was just thinking about this last week! I'd like a couple of small changes though.

flake.nix
BurNiinTRee: `callPackage` seems to fill in the `system` parameter automatically
```bash
$ nix repl nixpkgs
nix-repl> legacyPackages.x86_64-linux.callPackage ({system}: system) {}
"x86_64-linux"
```
on my system. So if I'm not missing anything w.r.t. `system` being something else when its coming from `flake-utils`, this shouldn't be an issue

YorikSar: Yes, you are correct. It basically ends up coming from line 65 here. No need to repeat it then! :)

BurNiinTRee: But since were on the topic of necessary arguments. is there a reason we can't just use `../.` inside `lib/lib.nix` instead of passing in `self.outPath`?

thufschmitt (approved):

YorikSar (approved):

PR #135, opened 2023-09-12
Provide a way of using Organist without flakes

YorikSar:
Some people (see https://github.com/nickel-lang/organist/pull/134 by @BurNiinTRee for example) want to use Organist without flakes.
This PR:
- documents a way for doing that and adds a test for it
- cleans up `lib.nix` arguments so that `pkgs.callPackage` can subsitute all of them
- cleans up `importNcl` interface so that it's easy to use with just a path to a directory

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,3 +39,30 @@ Organist can be used to declare the dependencies for your project.
 These can then be instantiated using [Nix](https://nixos.org/nix).
 
 More information on <./doc/dependency-management.md>.
+
+## Using without flakes
+
+While Organist is focused on usage with flakes, you can use it without them like this:
+
+```nix
+# shell.nix
+let
+  pkgs = import <nixpkgs> {};
+  organistSrc = builtins.fetchTarball "https://github.com/nickel-lang/organist/archive/main.tar.gz";
+  organist = pkgs.callPackage "${organistSrc}/lib/lib.nix" {inherit organistSrc;};
+in
+  (organist.importNcl {baseDir = ./.;}).shells.default
+```
+
+And then use `nix develop -f shell.nix` as usual. Note that this will not use nixpkgs and Nickel from Organist's flake.lock.
+You can use flake-compat to use them:
+
+```nix
+# shell.nix
+let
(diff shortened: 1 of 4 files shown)


YorikSar:
@BurNiinTRee I've changed `importNcl` interface here, so you'll have to change how you use it, but I think this way is cleaner. Please take a look.

BurNiinTRee:
It's great outside of flakes, but the `nickel-res.json` build fails when using it from a flake with a  `nickel.lock.ncl` present due to the imported path to `lib/lib.ncl` not being present in the sandbox.
see [this gist](https://gist.github.com/BurNiinTRee/d85499f53ef85bdb4649614f7377cbe8)

YorikSar:
@BurNiinTRee It looks like making this work without `flake-compat` would require passing the root of the Organist sources to `callPackage`, because `${../.}` that I'm using here creates a copy of sources in Nix store with different path, so lockfile will always be off. I've extracted the fix for running without `nickel.lock.ncl` to #137 so that we can merge it separately.

What do you think about using `flake-compat`?

YorikSar:
Depends on #137.

I've added back `flakeRoot` with name `organistSrc`, so user will have to either use flakes or provide fetched tarball, which should match Organist flake and thus keep matching lockfile

YorikSar:
Oh, for some reason shells without flakes don't really work now... All checks other than `bash --version` fail. Will investigate.

YorikSar:
Forgot `--command bash` for new `nix develop` lines. Should be all better now! 🎉

BurNiinTRee:
This seems to work in all really well. Thank you :)

I'm not sure if we misunderstood, but what wasn't working before was using it from a flake **with** the lockfile, it worked well outside of the flake with and without the lockfile and from a flake without or with an outdated lockfile.
But now everything i threw at it worked as expected

YorikSar:
@BurNiinTRee Before it wasn't working if lockfile on disk matched one that was calculated. For non-flake, it was calculating it wrong, so it wasn't failing if it existed. Now it should be calculating it in the same way for both flake and non-flake, and with #137 not fail what it matches.

thufschmitt (approved):

PR #138, opened 2023-09-17
Add shim for `builtins.placeholder` to lib.ncl

BurNiinTRee:
This is in line with `import_nix` or `import_file`.

This will be necessary if there will ever be a "organist-pkgs" independent of `nixpkgs`


thufschmitt (approved):
Thanks, that's indeed something that's bound to be useful

Issue #58, opened 2023-04-19
Generate 3d party config files from Nickel


thufschmitt:
A few potentially useful things:

- Generate pre-commit hooks (a la pre-commit-hooks.nix)
- Generate CI configs (like https://github.com/tweag/tf-ncl/commit/0b021022ca6b46ab3d75fba0e988a9f87f4765ce)

https://www.schemastore.org/json/ provides some schemas for a lot of potential targets, a compiler from these to Nickel contracts could be really helpful there

marijetamh:
This task is a requirement for the first release

thufschmitt:
Underway. The basic functionality is here, now need to clean it up a bit

PR #139, opened 2023-09-18
Dogfood devshells

YorikSar:
Add project.ncl and nickel.lock.ncl, use them for this very flake. For now use only devshells.

Update cachix/install-nix-action to install Nix 2.17 that doesn't evaluate `packages` for all systems on `nix flake check`.

Preparation for dogfooding future improvements from https://github.com/nickel-lang/organist/issues/58

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -18,7 +18,44 @@
     nixpkgs,
     flake-utils,
     nickel,
-  } @ inputs:
+  } @ inputs: let
+    # Generate typical flake outputs from .ncl files in path for provided systems (default from flake-utils):
+    #
+    # apps.${system}.regenerate-lockfile generated from optional lockFileContents argument,
+    #   defaulting to `organist` pointing to this flake
+    # devShells.${system} and packages.${system} generated from project.ncl
+    #
+    # (to be extended with more features later)
+    outputsFromNickel = baseDir: flakeInputs: {
+      systems ? flake-utils.lib.defaultSystems,
+      lockFileContents ? {
+        organist = "${self}/lib/nix.ncl";
+      },
+    }:
+      flake-utils.lib.eachSystem systems (system: let
+        lib = self.lib.${system};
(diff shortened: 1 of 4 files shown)


flake.nix
```diff
@@ -19,7 +19,7 @@
     flake-utils,
     nickel,
   } @ inputs:
-    {
+    builtins.foldl' (prev: next: prev // next prev) {
```
thufschmitt, line 22: This is quite hard to follow. Why can't we instead have something like below?

```nix
let
  outputsFromNickel = ...;
  # whatever else is required ...;
in
(outputsFromNickel ./. inputs
{
  lockFileContents.organist = "./lib/nix.ncl";
}) //
{
  inherit outputsFromNickel;
  # All the pure Nix outputs
}

File: flake.nix [hunk 2]
```diff
@@ -52,67 +52,74 @@
...
-              pkgs.parallel
-              pkgs.gnused
-            ];
-            text = builtins.readFile ./run-test.sh;
-          };
-        in {
-          type = "app";
-          program = pkgs.lib.getExe testScript;
-        };
+            apps =
```
thufschmitt, line 79: Is there a fundamental reason why we can't define the apps in Nickel, or is it just a temporary thing?

YorikSar:
I've added `foldl'` to be able to use things that are not available via `self` without causing infinite recursion like `outputsFromNickel` and `apps` that we need to override to add flake-specific apps. I'm working on a way to define apps in Nickel, so for now we need to keep them in flake.nix. The scope of this PR is to bootstrap dogfooding so that we can migrate more and more bits to Nickel later.

Since `foldl'` approach is confusing to people and doesn't really help on the very next step, I'll remove it and just fall back to bunch of `let`s.

thufschmitt (approved):
The `let` version is much more readable!

Looks good to go now, let's merge!

PR #140, opened 2023-09-25
Allow to specify some flake outputs from Nickel

YorikSar:
Add `flake` field to `project.ncl` that allows to specify certain flake outputs. For example, you can add lines like these to run `hello` as an app or a check:

```
flake.apps.hello.program = nix-s%"%{import_nix "nixpkgs#hello"}/bin/hello"%,
flake.checks.hello = import_nix "nixpkgs#hello",
```

I will be gragually migrating apps and checks from our `flake.nix` to `project.ncl` using this.


YorikSar:
@thufschmitt I’m not sure we want to expose everything at top level, so I’ve decided to namespace it. Mixing flake API with Organist one doesn’t feel right to me because it’s less declarative, less abstract and less typed. I feel like `flake` section should be an “output” of Nickel part.
 If you want, I can move it to the top level now, or we could do that at any point later if we decide that’s better.

thufschmitt (approved):

PR #141, opened 2023-09-25
Implement ShellApplication in Nickel

YorikSar:
Also move `run-tests` app from `flake.nix` to `project.ncl`.

Part of https://github.com/nickel-lang/organist/issues/58.

### DIFF
--- a/lib/builders.ncl
+++ b/lib/builders.ncl
@@ -8,6 +8,15 @@ let concat_strings_sep = fun sep values =>
   else
     std.array.reduce_left (fun acc value => nix-s%"%{acc}%{sep}%{value}"%) values
 in
+
+let MutExclusiveWith = fun other name other_name label value =>
+  if value == null && other == null then
+    std.fail_with "You must specify either %{name} or %{other_name} field"
+  else if value != null && other != null then
+    std.fail_with "You can set only one of %{name} or %{other_name} fields"
+  else
+    value
+in
 {
   NickelPkg
   # we should only need two '%%', but a current Nickel bug (#XXX) bug makes the example being
@@ -84,4 +93,49 @@ in
         structured_env.buildInputs = packages,
       }
         | (NickelPkg & { packages | { _ : Derivation } }),
+
+  ShellApplication
+    | doc m%"
+    "%
+    =
+      NixpkgsPkg
+      & {
+        name,
+        version | default = "0.0.0",
+        content.text | NixString | optional,
+        content.file | NixString | default = lib.to_file name content.text,
+        runtime_inputs | { _ : Derivation } = {},
+        shell_binary | NixString | default = lib.import_nix "nixpkgs#runtimeShell",
+
+        env.buildInputs.coreutils = lib.import_nix "nixpkgs#coreutils",
+        env.buildCommand = nix-s%"
+          mkdir -p "${outputs[out]}/bin"
+          target="${outputs[out]}/bin/${name}"
+          (cat <<<"$scriptHeader"; cat "%{content.file}") > "$target"
+          chmod +x "$target"
+          %{lib.import_nix "nixpkgs#stdenv.shellDryRun"} "$target"
+          "%{lib.import_nix "nixpkgs#shellcheck"}/bin/shellcheck" "$target"
+        "%,
+        env.scriptHeader =
+          let path_line =
+            if std.record.is_empty runtime_inputs then
+              ""
+            else
+              let paths =
+                runtime_inputs
+                |> std.record.values
+                |> std.array.map (fun s => nix-s%"%{s}/bin"%)
+                |> concat_strings_sep ":"
+              in
+              nix-s%"export PATH="%{paths}:$PATH""%
+          in
+          nix-s%"
+            #!%{shell_binary}
+            set -o errexit
+            set -o nounset
+            set -o pipefail
+            %{path_line}
+          "%,
+      }
+        | NickelPkg,
 }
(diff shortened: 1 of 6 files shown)


lib/builders.ncl
```diff
@@ -84,4 +93,75 @@ in
...
+        shell_binary | NixString | default = nix-s%"%{lib.import_nix "nixpkgs#bash"}/bin/bash"%,
+
+        build_command = {
+          cmd = nix-s%"%{lib.import_nix "nixpkgs#bash"}/bin/bash"%,
+          args = [
+            "-c",
+            nix-s%"
+              set -euo pipefail
+              source .attrs.sh
+              PATH="%{lib.import_nix "nixpkgs#coreutils"}/bin:$PATH"
```
thufschmitt, line 116: Why not add `coreutils` to `nativeBuildInputs`?

lib/builders.ncl
```diff
@@ -84,4 +93,75 @@ in
...
+                else
+                m%"(cat <<<"$scriptHeader"; cat "$scriptPath") > "$target""%
+              }
+              chmod +x "$target"
+              "%{shell_binary}" -n -O extglob "$target"
+              "%{lib.import_nix "nixpkgs#shellcheck"}/bin/shellcheck" "$target"
+            "%
+          ],
+        },
+        nix_drv =
```
thufschmitt, line 130:
```suggestion
        env =
```

lib/builders.ncl
```diff
@@ -84,4 +93,75 @@ in
...
+
+  ShellApplication
+    | doc m%"
+    "%
+    =
+      {
+        name,
+        version | default = "0.0.0",
+        text | NullOr NixString | MutExclusiveWith file "text" "file" | default = null,
+        file | NullOr String | default = null,
```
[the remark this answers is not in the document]
thufschmitt: Actually, another more idiomatic solution might be

```nickel
{
  content.text | optional | NixString,
  content.file | NixString | default = lib.to_file content.text,
}
```

(where `lib.to_file` has to be implemented, but it's something quite useful any way)

thufschmitt (changes requested):
Some minor suggestions, but looks good otherwise, except for the lack of tests. We should think about an easy way to write tests for this kind of builders :thinking:

lib/builders.ncl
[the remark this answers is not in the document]
YorikSar: > `TaggedUnion`

Is it builtin or implemented somewhere? I'm not sure how to implement it preserving record contract semantics for LSPs...

> `lib.to_file`

I actually wanted to avoid creation of almost duplicate paths in /nix/store if possible, so if the text is available in Nickel, I'd prefer to write it to .drv rather than into separate file. Although I might be convinced otherwise...

lib/builders.ncl
YorikSar: `nativeBuildInputs` is handled by `stdenv`, this derivation doesn't use it.

lib/builders.ncl
YorikSar: `env` is handled by `NixpkgsPkg`, this is `NickelPkg`.

YorikSar (commented):
> the lack of tests

Isn't `run-test` implementation good enough as a test here? Of course it doesn't cover some branches in code, but it feels like it should be enough.

I could try implementing something like a "golden master" tests to be able to quickly cover more branches, but I think it should be a separate effort.

lib/builders.ncl
thufschmitt: Oh indeed. Why?

lib/builders.ncl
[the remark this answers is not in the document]
thufschmitt: What's the difference between having it in the drv and in a separate file?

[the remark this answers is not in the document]
thufschmitt: > Is it builtin or implemented somewhere? I'm not sure how to implement it preserving record contract semantics for LSPs...

It's indeed not implemented. But it could reuse the same logic as `MutExclusiveWith`, just with one layer of nesting (not pretty, but that would work)

lib/builders.ncl
YorikSar: `stdenv` doesn't provide much benefit here, so why depend on it if we can avoid it?

lib/builders.ncl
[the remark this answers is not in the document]
thufschmitt: > we won't be able to enforce setting either text or file fields with contract.

No, but that's actually probably required for overriding to work properly. We probably want something like the above to work:

```nickel
let myApp = {
  name = "foo",
  content.file = "foo.sh",
} in
myApp & {
  content.text = "echo Overriden"
}
```

(ideally we would have that if `content.file` is defined, then `content.text` has to be `force`-ed to be overridden, but I don't think that's possible)

lib/builders.ncl
thufschmitt: Well, it's easier and more standard. So I'd rather use it everywhere for the sake of simplicity

thufschmitt:
@YorikSar I've added a test for it, let me know what you think

thufschmitt (approved):

YorikSar:
@thufschmitt I got stuck on this error here:

```
     ┌─ /nix/store/0bfh4kzbmb2dn9rh4fx977zbzqvxgbb6-r0fb3am31yn4kgsds9z58ckdzxrdfr7n-source/lib/builders.ncl:106:63
     │
 106 │         content.file | NixString | default = lib.to_file name content.text,
     │                                                               ^^^^^^^^^^^^ this requires the field `text` to exist
```

I'm not sure where it comes from and how. `(import "project.ncl").flake.checks` works, but `(import "project.ncl").flake` throws this error. I'll continue trying to figure it out on Monday.

I've also changed tests a bit, please check if it's ok.

PR #143, opened 2023-09-28
Use Nickel from Nixpkgs rather than its own flake

thufschmitt:
This reduces the Nix evaluation time by a non-trivial factor (1.6x
faster according to hyperfine). Probably because Nickel's own flake uses
Crane (which does some fancy eval-time magic) while the Nixpkgs version
uses the more brutal (but less eval-costly) `buildRustPackage`.

```console
$ hyperfine --warmup 1 -L dir nickel-from-flake,nickel-from-nixpkgs "nix develop --option eval-cache false ./{dir} --command true"
Benchmark 1: nix develop --option eval-cache false ./nickel-from-flake --command true
  Time (mean ± σ):     17.661 s ±  0.053 s    [User: 12.093 s, System: 2.711 s]
  Range (min … max):   17.579 s … 17.739 s    10 runs

Benchmark 2: nix develop --option eval-cache false ./nickel-from-nixpkgs --command true
  Time (mean ± σ):     11.059 s ±  0.036 s    [User: 7.265 s, System: 1.829 s]
  Range (min … max):   11.007 s … 11.140 s    10 runs

Summary
  'nix develop --option eval-cache false ./nickel-from-nixpkgs --command true' ran
    1.60 ± 0.01 times faster than 'nix develop --option eval-cache false ./nickel-from-flake --command true'
```


project.ncl
```diff
@@ -7,7 +7,7 @@ let import_nix = organist.lib.import_nix in
     organist.shells.Bash
     & {
       build.packages = {
-        nickel-lang-cli = import_nix "nickel#nickel-lang-cli",
+        nickel-lang-cli = import_nix "nixpkgs#nickel",
```
YorikSar, line 10:
```suggestion
        nickel = import_nix "nixpkgs#nickel",
```

I guess, there's not point in keeping the name from nickel flake.

project.ncl
```diff
@@ -7,7 +7,7 @@ let import_nix = organist.lib.import_nix in
...
         parallel = import_nix "nixpkgs#parallel",
         gnused = import_nix "nixpkgs#gnused",
       },
```
YorikSar, line 13: Also, https://github.com/nickel-lang/organist/pull/142 added another reference to `nickel#lsp-nls` here.

YorikSar (changes requested):
Looks like there's another reference to `nickel` input in https://github.com/nickel-lang/organist/pull/143/files#diff-206b9ce276ab5971a2489d75eb1b12999d4bf3843b7988cbe8d687cfde61dea0L79.

thufschmitt:
Thanks for the review @YorikSar. I've rebased it on top of main and (hopefully) answered your remarks. PTAL

thufschmitt:
We're already kind-of doing that, so closing

PR #151, opened 2023-09-29
Namespace the Nix-related things

thufschmitt:
Organist isn't _just_ a Nix framework, so we want to make sure that it
is possible to add stuff that's not strictly Nix-related to it, and have
it feel first-class.
So move most of the Nix-related stuff in its own namespace to clear up
some real estate for the rest.

Yet another step toward #58


examples/c-hello-world/project.ncl
```diff
@@ -4,21 +4,21 @@ let organist = inputs.organist in
...
-    organist.builders.NixpkgsPkg
+    organist.nix.builders.NixpkgsPkg
     & {
       name = "hello",
       version = "0.1",
       nix_drv = {
-        buildInputs.gcc = organist.lib.import_nix "nixpkgs#gcc",
-        buildInputs.coreutils = organist.lib.import_nix "nixpkgs#coreutils",
-        buildInputs.bash = organist.lib.import_nix "nixpkgs#bash",
+        buildInputs.gcc = organist.nix.builtins.import_nix "nixpkgs#gcc",
```
aspiwack, line 12: This is a lot of boilerplate (it was before too, I must say). Note how the list says “Nix” 3 times! Is this something that we expect to be often part of an Organist project description?

thufschmitt: That's indeed quite wordy for something that's bound to be rather common. Let me re-export that from the toplevel of organist, it should already make it more tractable

PR #152, opened 2023-09-29
Move checks to project.ncl

YorikSar:
It's unclear if we need to implement runCommand ourselves, it's quite simple to achieve necesary result with plaing NixpkgsPkg override.

Introduce NixEnvironmentVariable contract to handle all possible things than might end up in `env`.

Part of https://github.com/nickel-lang/organist/issues/58.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -100,31 +100,6 @@
               program = pkgs.lib.getExe testScript;
             };
           };
-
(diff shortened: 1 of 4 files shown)


thufschmitt:
Yes you're right, let's do that for the time being

thufschmitt (approved):

Issue #144, opened 2023-09-28
Add a mechanism for generating files in the repo


thufschmitt:
Add a `files` field to the toplevel Nickel schema for defining files that should be generated in-place in the repo.

For instance:

```
{
  files.myConfigFile = {
    destination = "myConfigFile", # optional, defaults to the name of the field
    content.text = m%"
      foo = 1,
      bar = "baz"
    "%,
  },
}
```

This could be triggered by calling a specific `app` in the flake (`nix run .#organist regenerate` or something like that).

Eventually, this could also replace the current `regenerate-lock-file` logic

YorikSar:
What would be the use case for `destination` field? You can specify anything in the name, can't you?

thufschmitt:
Yes, that's mostly for overriding purposes. If for some reason you want to change the target, you can override the `destination` field (while overriding the field name isn't really possible). Random example of such a thing from my past self: https://github.com/nix-community/home-manager/issues/2360#issuecomment-1009723929

PR #155, opened 2023-10-13
Add a crude file generation mechanism

thufschmitt:
Running `nix run .\\#regenerate-files` now generates files locally
according to the `files` option in `project.ncl`.

This is still rather crude in a number of ways, but it is already
working (and used to generate this project's own `.gitignore`).

Fixes https://github.com/nickel-lang/organist/issues/144


project.ncl
```diff
@@ -73,5 +73,11 @@ let import_nix = organist.nix.import_nix in
     },
 
   flake.checks = import "tests/main.ncl",
+
+  files.".gitignore".materialisation_method = 'Copy,
+  files.".gitignore".content = m%"
+    /examples/*/result
+    /result
+  "%,
```
YorikSar, line 81:
```suggestion
    /result

  "%,
```

We should have a newline at the end there. We could also have a separate field like `add_newline` that would by default add a newline at the end of file.

lib/files.ncl
```diff
@@ -0,0 +1,63 @@
...
+          let copy_command =
+            match {
+              'Symlink => "ln -s",
+              'Copy => "cp",
+            }
+              file_descr.materialisation_method
+          in
+          let file_in_store = nix.builtins.to_file key file_content in
+          nix-s%"
+            (chmod +w %{target}; rm %{target}) || true
```
YorikSar, line 42:
```suggestion
            rm -f %{target}
```

`chmod` requires rights to modify the inode, but we only need to remove the file from the directory. Also, `rm -f` doesn't complain about access rights and missing files.

.gitignore
```diff
@@ -1,2 +1,2 @@
 /examples/*/result
```
YorikSar, line 1: We should also add a comment at the top mentioning that this file is autogenerated. Maybe add to File attribute `preamble` defaulting to something like `# Autogenerated file, do not edit`, it should fit majority of formats that we want to support and leave an escape hatch for those who need to override it.

This could probably be done in a separate PR though.

lib/files.ncl
```diff
@@ -0,0 +1,63 @@
...
+      | default
+      = null,
+    content
+      | doc m%"
+        The content of the file.
+      "%
+      | nix.derivation.NixString,
+    materialisation_method : [| 'Symlink, 'Copy |]
+      | doc m%"
+        How the file should be materialized on-disk.
```
YorikSar, line 19: I think for our primary use case we care more about representation in Git repo, not on disk.

lib/schema.ncl
```diff
@@ -24,7 +25,16 @@ let nix = import "./nix-interop/nix.ncl" in
...
-    flake | FlakeOutputs | optional,
+    files
+      | filegen.Files
+      | doc m%"
+        Set of files that should be generated in the project's directory.
+      "%
+      | default
+      = {},
+    flake | FlakeOutputs = {
+        apps.regenerate-files.program = nix-s%"%{filegen.regenerate_files files}/bin/regenerate-files"%,
```
YorikSar, line 36: Should probably also add a check that would verify that files are properly generated here.

YorikSar (changes requested):

lib/files.ncl
thufschmitt: Indeed, thanks

lib/files.ncl
```diff
@@ -0,0 +1,63 @@
...
+
+        Symlinking is generally preferred as it makes it easier to track where
+        the files are coming from, but their target only exists after a first
+        call to Organist, which might be undesirable.
+      "%
+      | default = 'Symlink,
```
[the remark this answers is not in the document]
thufschmitt: That makes sense. Maybe we could rename that to a switch between `persistent` or not in the future

lib/schema.ncl
thufschmitt: Indeed. Let me save that for a follow-up PR

lib/files.ncl
thufschmitt: It is, but I'll save these considerations for when we mention `.gitignore` as it'll be more relevant there (and we don't know anything about Git yet)

.gitignore
thufschmitt: I've explicitly added that to the text snippet for now

thufschmitt:
@YorikSar I think I've answered your comments, let me know what you think

lib/files.ncl
YorikSar: Good point. Yes, we should keep it somewhat consistent.

.gitignore
YorikSar: Should work for this file for now. Could you please update this file though? It seems it doesn't match the definition now.

YorikSar (changes requested):
Just one small issue and then gtg :)

.gitignore
thufschmitt: Oh indeed, thanks. Fixed

PR #158, opened 2023-10-18
Add doc to ShellApplication

YorikSar:
Found in local stash, should be useful.


thufschmitt (approved):
Quite handy indeed :)

PR #161, opened 2023-10-26
Update Nickel

thufschmitt:
With https://github.com/tweag/nickel/pull/1699 fixed, we can now use the latest Nickel and benefit from all the coolness that come with it :)


thufschmitt:
Let's not merge right now as it's causing a huge increase of the evaluation time.

Since this is very likely due to Nickel's own flake evaluation time (because of Crane), I'd better wait for a new release and then take Nickel from Nixpkgs (which doesn't suffer from this huge eval time)

PR #174, opened 2023-11-09
Add the `packages` of the shells to `nativeBuildInputs`

### LINKED ISSUE #173: when added as packages in a shell, pkg-config doesn't find packages
With the `project.ncl` below, if I enter the shell and type `pkg-config --cflags openssl`, it complains that "Package openssl was not found in the pkg-config search path."

```nickel
let inputs = import "./nickel.lock.ncl" in
let organist = inputs.organist in

{
  shells = organist.shells.Bash,

  shells.build = {
    packages = {},
  },

  shells.dev = {
    packages.openssl = organist.import_nix "nixpkgs#openssl",
    packages.pkg-config = organist.import_nix "nixpkgs#pkg-config",
  },
}
  | organist.OrganistExpression
```

thufschmitt:
`nativeBuildInputs` is what we're interested in by default. In particular, this is needed to trigger the `cc` or `pkg-config` setup hooks which populate their respective search path environment variables automatically.

This is also what the Nixpkgs `mkShell` function does (probably for the same reason).

Fixes #173

### DIFF
--- a/examples/c-hello-world/hello.c
+++ b/examples/c-hello-world/hello.c
@@ -1,6 +1,6 @@
 #include <stdio.h>
+#include "hello.h"
 
-int main() {
+void hello() {
     printf("Hello, world!");
-    return 0;
 }
--- a/examples/c-hello-world/hello.h
+++ b/examples/c-hello-world/hello.h
@@ -1,0 +1,1 @@
+void hello();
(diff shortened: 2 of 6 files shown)


thufschmitt:
@jneeman I'm writing a test for this, but can you confirm that it fixes #173 for you?

Issue #179, opened 2023-11-26
Fancier rust shells


jneem:
Organist's rust shell uses the rust toolchain from nixpkgs, which is usually fine but lacks a few options. In particular, I often want to
- use nightly,
- add cross-compilation targets,
- add extra components (usually miri, but sometimes also rust-src)

All these options (and more) are supported by nix-community/fenix, so maybe a nice organist interface could be built on top of that?

I had a quick stab at it and got something working, which I will add as an attachment. It requires modifying the template's flake.nix to pass the fenix input like
```
    organist.flake.outputsFromNickel ./. (inputs // { fenix = inputs.fenix.packages.${system}; }) {};
```
because I couldn't figure out how else to thread the system through...

[project.ncl.txt](https://github.com/nickel-lang/organist/files/13465967/project.ncl.txt)

thufschmitt:
Nice, thanks!

I'll look at how this can be upstreamed.

Some quick thoughts:

- You shouldn't need the `{ fenix = inputs.fenix.packages.${system}; }` trick in the Flake because `import_nix` implicitly tries adding `packages.${system}` to the attribute path you pass to it. It's a bug if that doesn't work
- Maybe the components arrays could be replaced by a record of booleans, something like:
    ```nickel
      enabledComponents | "Components to install" = {
        rustc | bool | default = true,
        cargo | bool | default = true,
        rust-std | bool | default = true,
        rustfmt | bool | default = false,
        rust-docs | bool | default = false,
        rust-analyzer | bool | default = false,
        clippy | bool | default = false,
        miri | bool | default = false,
        rust-src | bool | default = false,
      }
    ```

    That's a bit more verbose, but plays nicer with the merge semantics.

jneem:
> You shouldn't need the { fenix = inputs.fenix.packages.${system}; } trick in the Flake

Indeed I don't, thanks! I was so convinced that threading through the `system` was going to be annoying that I didn't think to check whether it worked without it...

> Maybe the components arrays could be replaced by a record of booleans

Yes, that works nicely. I even did the same to extraTargets, which is nice with autocomplete because the target names are long. I put the result [here](https://github.com/jneem/organist-rust), since there are now a few files involved.

thufschmitt:
Fixed by #180

PR #180, opened 2023-12-06
A fancier rust shell based on fenix.

jneem:
Here's an attempt at upstreaming the rust shell in #179.

I'm not thrilled about the way it handles extra targets -- I generated the list from fenix, but I couldn't find a way to automatically keep it up-to-date. I think the current state is somewhat workable, though, because (1) the targets don't change too much and (2) even if your favorite target is missing from the predefined list of targets, you can still set it and it will work. The predefined list only affects autocompletion.


thufschmitt (approved):
Thanks (and sorry for the huge delay).

Agreed, the hardcoded list of targets isn't great, but we can revisit that later

thufschmitt:
(push-forced to re-trigger the CI which was stuck for some reason)

jneem:
Looks like the template test assumes that the field name in `packages` is the same as the name of some binary that gets installed. I renamed the package from `rust-toolchain` to `rustc` and it now passes (but it cancelled the automerge)

PR #182, opened 2023-12-16
direnv: allow adding additional files to watch

nrdxp:


lib/direnv.ncl
```diff
@@ -16,7 +23,7 @@
             if ! has nix_direnv_version || ! nix_direnv_version 2.4.0; then
               source_url "https://raw.githubusercontent.com/nix-community/nix-direnv/2.4.0/direnvrc" "sha256-XQzUAvL6pysIJnRJyR7uVpmUSZfc7LSgWQwq/4mBr1U="
             fi
-            nix_direnv_watch_file project.ncl nickel.lock.ncl
+            nix_direnv_watch_file project.ncl nickel.lock.ncl %{std.string.join " " direnv.watch_files}
```
yannham, line 26: Shouldn't we escape the file names here, in all generality? For example:

```nickel
let escape = std.string.replace "\"" "\\\"" in
let enclose = fun s => "\"%{s}\"" in
let extra_watch_files =
  direnv.watch_files
  |> std.array.map (fun s => s |> escape |> enclose)
  |> std.string.join " "
in
..
            nix_direnv_watch_file project.ncl nickel.lock.ncl extra_watch_files
```

nrdxp: I'm assuming this is for ` ` and other special characters? Perhaps we should use single quotes in the `enclose` function to also avoid arbitrary command expansions?

yannham: Right, `'` is probably better (then we need to escape them instead of `"` in `escape`)

yannham: @nrdxp are you still willing to apply this change? I think after that this PR is good to go!

PR #190, opened 2024-03-21
Fix tests on Nix 2.19+

YorikSar:
Because of https://github.com/NixOS/nix/issues/9330 that probably will never be fixed, and https://github.com/NixOS/nix/issues/2678 that still might, we can't rely on passing JSON via stdin anymore. Use a temporary file instead.


thufschmitt (approved):
Thanks for that :)

PR #196, opened 2024-04-21
chore(templates): Use fully qualified nixpkgs URL

srid:
This is necessary to be able to use this flake on systems with empty Nix registry.


YorikSar (changes requested):
Please update `flake.lock` accordingly.

srid:
@YorikSar Which `flake.lock`? The template [does not have one](https://github.com/srid/organist/tree/patch-1/templates/default).

YorikSar (approved):
Ah, good point, I misunderstood which flake is involved here.

thufschmitt (approved):
Thanks

PR #198, opened 2024-04-24
Add nls to the default shell

thufschmitt:
It doesn't cost much since we need nickel any way, and it's a large part of the value of Organist

### DIFF
--- a/project.ncl
+++ b/project.ncl
@@ -7,7 +7,7 @@ let import_nix = organist.nix.import_nix in
     organist.shells.Bash
     & {
       build.packages = {
-        nickel-lang-cli = import_nix "nixpkgs#nickel",
+        nickel = import_nix "nixpkgs#nickel",
         parallel = import_nix "nixpkgs#parallel",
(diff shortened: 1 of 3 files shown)


project.ncl
```diff
@@ -7,7 +7,6 @@ let import_nix = organist.nix.import_nix in
     organist.shells.Bash
     & {
       build.packages = {
-        nickel-lang-cli = import_nix "nixpkgs#nickel",
```
yannham, line 10: I don't know if it's related, but the CI is failing (with a not very informative error). How do build packages and flake packages interact? Is it ok to remove this one?

thufschmitt: Mh, maybe not.

But that's also not the reason for the CI failure. The CI just complains that `nls` doesn't accpet a `--version` flag :D

```
+ nls --version
error: unexpected argument '--version' found
```

yannham: Argh, this is embarrassing.

yannham (approved):

PR #199, opened 2024-04-25
Allow codeblocks to be copied and pasted

ShalokShalom:
This commit allows the codeblocks to be actually copied and pasted


ShalokShalom:
I would recommend, to always put the output into a separate box.

PR #203, opened 2024-05-27
Make the file generation more transparent

thufschmitt:
Regenerate the files (defined in `files.*`) automatically when entering the shell (if needed).

Also unify that mechanism with the one used to generate the lockfile so that everything is generated at once.

The lockfile generation is slightly hacky because there's no easy way to pass its content from Nix to Nickel. But that'll have to do the job.


lib/lib.nix
```diff
@@ -210,9 +206,14 @@
     nickelResult = callNickel {
       inherit nickelFile baseDir flakeInputs lockFileContents;
     };
+    enrichedFlakeInputs =
+      flakeInputs
+      // {
+        "%%organist_internal".nickelLock = builtins.toFile "nickel.lock.ncl" (buildLockFileContents lockFileContents);
```
yannham, line 212: Could be nice to add a little comment explaining why we need to do that, for future you, me and others

run-test.sh
```diff
@@ -120,7 +119,7 @@ test_example () (
   cp -r "$examplePath" ./example
   pushd ./example
   prepare_shell
-  nix run .\#regenerate-files --print-build-logs
+  # nix run .\#regenerate-files --print-build-logs
```
yannham, line 122: Is this line still there on purpose or should it just be scraped entirely?

yannham (approved):

run-test.sh
thufschmitt: Nope, that's purely accidental. Removed it

lib/lib.nix
thufschmitt: Indeed. Fixed

thufschmitt:
It turns out that this blows-up the Nickel evaluation time (~x30 according to my benchmarks). I'll have to investigate why before we can merge this

yannham:
x30 for such a change is really suspicious...

thufschmitt:
> x30 for such a change is really suspicious...

https://github.com/tweag/nickel/issues/1930 seems to be the culprit (possibly also the real culprit for https://github.com/tweag/nickel/issues/1630)

thufschmitt:
Rebased on top of #205 . Somehow, this sidesteps the performance issue :partying_face:

thufschmitt:
(Note that this accidentally includes https://github.com/nickel-lang/organist/pull/210. I can revert it if you need @yannham )

lib/lib.nix
```diff
@@ -169,23 +169,19 @@
...
-          ${nickel}/bin/nickel export eval.ncl --field config.flake > $out
-        ''
+      then ''
+        cp -r "${sources}" sources
+        if [ -f sources/nickel.lock.ncl ]; then
+          chmod +w sources sources/nickel.lock.ncl
+        else
+          chmod +w sources
+        fi
+        cp $expectedLockfileContentsPath sources/nickel.lock.ncl
```
yannham, line 179:
```suggestion
        cp "$expectedLockfileContentsPath" sources/nickel.lock.ncl
```

lib/files.ncl
```diff
@@ -43,17 +53,17 @@ let regenerate_files | Files -> nix.derivation.Derivation = fun files_to_generat
...
         nix-s%"
-            rm -f %{target}
-            echo "Regenerating %{target}"
-	    target_dir=$(dirname %{target})
-	    test "${target_dir}" != "." && mkdir -p "${target_dir}"
-            %{copy_command} %{file_in_store} %{target}
+            if [[ ! -f "%{target}" ]] || [[ $(cat "%{target}") != $(cat "%{file_descr.file}") ]]; then
+              rm -f %{target}
+              echo "Regenerating %{target}"
+              target_dir=$(dirname %{target})
```
yannham, line 60:
```suggestion
              target_dir=$(dirname "%{target}")
```

lib/files.ncl
```diff
@@ -43,17 +53,17 @@ let regenerate_files | Files -> nix.derivation.Derivation = fun files_to_generat
...
+              test "${target_dir}" != "." && mkdir -p "${target_dir}"
+              # XXX: If `source.file` is set explicitely to a relative path
+              # and `materialisation_method` is `'Symlink`, this will link to the
+              # original file, not one in the store. Not sure that's what we want.
+              %{copy_command} %{file_descr.file} %{target}
```
yannham, line 65:
```suggestion
              %{copy_command} "%{file_descr.file}" "%{target}"
```

lib/files.ncl
```diff
@@ -83,10 +93,31 @@ let regenerate_files | Files -> nix.derivation.Derivation = fun files_to_generat
...
+      filegen_hook,
+      shells,
       flake.apps.regenerate-files.program = nix-s%"%{regenerate_files files}/bin/regenerate-files"%,
+
+      shells.build.hooks =
+        if filegen_hook.enable then
+          {
+            filegen_hook = nix-s%"
+            %{regenerate_files files}/bin/regenerate-files
+          "%
```
yannham, line 118: The indentation is a bit messed up.
```suggestion
            filegen_hook = nix-s%"
              %{regenerate_files files}/bin/regenerate-files
            "%
```

lib/lib.nix
yannham: Has the doc been lost in a force push?I can't see it still

yannham (approved):

PR #205, opened 2024-06-10
Use a module system for the different integrations

thufschmitt:
Uses a module-system (inspired by the NixOS one) to make the whole thing a cleaner:

1. Clearly separates the options provided by the modules from their implementations
2. Actually use merging for combining modules rather than abusing contract application (avoids issues like https://github.com/tweag/nickel/issues/1930)
3. Gives a standard way of writing modules so that they can be unified rather than each having a different ad-hoc structure.


lib/modules/main.ncl
```diff
@@ -0,0 +1,44 @@
+# A library defining some kind of module system for Nickel
+{
+  T
+    | doc m%"
+      The type of a module.
+    "%
+    = {
+      config | Schema = {},
+      Schema
+        | not_exported
```
yannham, line 10:
```suggestion
        | {..}
        | not_exported
```

Should we add this little check that `Schema` is always a record?

lib/modules/main.ncl
```diff
@@ -0,0 +1,44 @@
...
+
+        => {
+          Schema = { mod1.foo | Number, mod2.foo | String },
+          config = { mod1.foo = 1, mod2.bar = "bar" }
+        }
+      ```
+    "%
+    = fun modules =>
+      let configs = modules |> std.record.map (fun k v => v.config) in
+      let schemas = modules |> std.record.map (fun k v => v.Schema) in
```
yannham, line 35:
```suggestion
      let configs = modules |> std.record.map (fun _ v => v.config) in
      let schemas = modules |> std.record.map (fun _ v => v.Schema) in
```

If you're using 1.7 yo can do that now :slightly_smiling_face:

lib/modules/main.ncl
```diff
@@ -0,0 +1,44 @@
...
+      Schema
+        | not_exported
+        = {},
+      "$__organist_type" | force = "module",
+    },
+
+  submodules
+    | { _ : T } -> T
+    | doc m%"
+      Joins all the given module into a single one, prefixing all the keys of
```
yannham, line 18:
```suggestion
      Joins all the given modules into a single one, prefixing all the keys of
```

lib/modules/main.ncl
```diff
@@ -0,0 +1,44 @@
...
+      their options with the key of the corresponding module.
+
+      ```nickel
```
yannham, line 21:
```suggestion

      # Example

      ```nickel
```

lib/modules/main.ncl
```diff
@@ -0,0 +1,44 @@
...
+      let configs = modules |> std.record.map (fun k v => v.config) in
+      let schemas = modules |> std.record.map (fun k v => v.Schema) in
+      {
+        Schema = schemas,
+        config | Schema = configs,
+      } | T,
+
+  merge2
+    | T -> T -> T
+    = fun m1 m2 => m1 & m2,
```
yannham, line 43: You can use the curried version `(&)` for that.

yannham: (and  `std.record.merge_all` to merge a list of records)

yannham (approved):
LGTM! The migration seems rather straightforward.

lib/modules/main.ncl
thufschmitt: Removed in b152d7af2a8e2da5bfec74ed05fef775f9cadc04

PR #208, opened 2024-06-17
Only export the `flake` field of the nickel config

thufschmitt:
Makes it clear what Nix has (or not) access to.

Also makes it possible in the future to fearlessly use enums in the config, as they won't be exported

Depends on #205

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -44,17 +44,15 @@
         then {}
         else {
           inherit nickelOutputs;
-          packages = nickelOutputs.packages or {} // nickelOutputs.flake.packages or {};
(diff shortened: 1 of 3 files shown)


yannham (approved):

YorikSar (approved):

YorikSar:
@thufschmitt sorry, I thought it was targeting master. Should I revert it?

thufschmitt:
> @thufschmitt sorry, I thought it was targeting master. Should I revert it?

Nah, don't bother. #205 is nearly ready any way

PR #209, opened 2024-06-17
Split up the Nix interop

thufschmitt:
Separate the bit about calling Nickel (`importNickel`), and the one really about Organist (`importOrganist`).

Doesn't change anything semantically, but

1. Makes the flow (hopefully) clearer
2. Allow exporting a lower-level `importNickel` function that just calls Nickel and imports back the result without the Organist magic


yannham (approved):
I think there's an `importNcl` left somewhere in `lib/lib.nix` within a comment; a search and replace might get rid of those.

Otherwise, I know it's provided as it is, but maybe a bit more documentation on what `importNickel` is doing so that non-organist users can make something out of it could be nice. But doesn't have to be in this PR, even if it can.

thufschmitt:
> I think there's an `importNcl` left somewhere in `lib/lib.nix` within a comment; a search and replace might get rid of those.

Indeed, a quick grep found a few extras in the documentation, thanks.

> Otherwise, I know it's provided as it is, but maybe a bit more documentation on what `importNickel` is doing so that non-organist users can make something out of it could be nice. But doesn't have to be in this PR, even if it can.

That's fair. I've added a quick blurb to the function. I don't think it deserves much right now (it's still a bit coupled with Organist, so not producion-ready as an independent thing).

PR #219, opened 2024-06-26
Make the file generation evaluation faster

thufschmitt:
Various improvements to the file generation that together bring some 1.15x speedup to the Nickel evaluation

### DIFF
--- a/lib/nix-interop/derivation.ncl
+++ b/lib/nix-interop/derivation.ncl
@@ -1,54 +1,12 @@
 let type_field = "$__organist_type" in
 
+let nix_string = import "./nix-string.ncl" in
+
 let predicate | doc "Various predicates used to define contracts"
-  = {
-    is_nix_path = fun x =>
-      std.is_record x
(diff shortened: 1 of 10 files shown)


lib/nix-interop/derivation.ncl
```diff
@@ -1,54 +1,12 @@
...
-      && value."%{type_field}" == "callNix",
-    is_string_fragment = fun x =>
-      is_derivation x
-      || std.is_string x
-      || is_nix_path x
-      || is_nix_placeholder x
-      || is_nix_to_file x
-      || is_nix_call x
-  }
+  = {}
```
yannham, line 6: Why do we keep an empty `predicates` here?

yannham (approved):

lib/nix-interop/derivation.ncl
thufschmitt: Because I forgot to drop it? :sweat_smile: 

Removed in 2fd0bd3

PR #222, opened 2024-10-25
Add `File.target` validation and escaping

Chan-Siu-Man:
This change adds these in `files.ncl`:
- `File.target` validation with `std.string.NonEmpty`, `RelativePath`, `NoParentTraversal`,
- `File.target` shell-escaping in `regenerate_one`.

Before this change, the `files` field has `File` with possibly pathological `target | String`, e.g.:
- Empty target, i.e., `""`,
- Absolute target, e.g., `"/etc/passwd"`,
- Parent traversals, e.g,. `"../../../../../../../etc/passwd"`.

This change adds schemas to eliminate such cases.

Also, in `regenerate_one`, we add `shell_escape` to `file_descr.target`:
```
nix-s%"regenerate_function '%{copy_command}' '%{file_descr.file}' $'%{shell_escape file_descr.target}'"%
```

By quoting with `$'…'` (dollar-sign and *single*-quotes) instead of `"…"` (*double*-quotes), this correctly handles paths with characters such as `"`, `'`, `*`, `\`, `\n` (newline), `\t` (tabs), `$`.

### DIFF
--- a/lib/files.ncl
+++ b/lib/files.ncl
@@ -15,20 +15,15 @@ let NoTraversal =
       'Ok
   )
 in
-let
-# A ValidTarget eliminates these pathological targets:
-# - Empty target, i.e., "",
-# - Absolute target, e.g., `"/etc/passwd"`,
-# - Parent traversals, e.g,. `"../../../../../../../etc/passwd"`.
-ValidTarget = std.contract.all_of [std.string.NonEmpty, RelativePath, NoTraversal]
-in
 let File = {
   target
     | doc m%"
         The file to write to.
         If null, defaults to the attribute name of the file.
       "%
-    | ValidTarget
+    | std.string.NonEmpty # avoids ""
+    | RelativePath # avoids "/etc/passwd"
+    | NoTraversal # avoids "../../../../../../../etc/passwd"
     | optional,
   content
     | doc m%"
--- a/lib/files.ncl
+++ b/lib/files.ncl
@@ -3,8 +3,10 @@ let RelativePath =
   std.contract.from_predicate (fun x => !(x |> std.string.characters |> std.array.at 0 == "/"))
 in
 let NoTraversal =
-  std.contract.from_predicate (fun x => !(x |> std.string.split "/" |> std.array.any (fun part => part == ".."))
-  )
+  std.contract.from_predicate
+    (
+      fun x => !(x |> std.string.split "/" |> std.array.any (fun part => part == ".."))
+    )
 in
 let File = {
   target
--- a/lib/files.ncl
+++ b/lib/files.ncl
@@ -1,18 +1,9 @@
 let nix = import "./nix-interop/nix.ncl" in
 let RelativePath =
-  std.contract.from_validator (fun x =>
-    if x |> std.string.characters |> std.array.at 0 == "/" then
-      'Error { message = m%"Absolute Path: Expecting a relative path, got an absolute path starting with "/"."% }
-    else
-      'Ok
-  )
+  std.contract.from_predicate (fun x => !(x |> std.string.characters |> std.array.at 0 == "/"))
 in
 let NoTraversal =
-  std.contract.from_validator (fun x =>
-    if x |> std.string.split "/" |> std.array.any (fun part => part == "..") then
-      'Error { message = m%"Invalid Traversal: Path contains ".."."% }
-    else
-      'Ok
+  std.contract.from_predicate (fun x => !(x |> std.string.split "/" |> std.array.any (fun part => part == ".."))
   )
 in
 let File = {
--- a/lib/files.ncl
+++ b/lib/files.ncl
@@ -1,11 +1,34 @@
 let nix = import "./nix-interop/nix.ncl" in
+let RelativePath =
+  std.contract.from_validator (fun x =>
+    if x |> std.string.characters |> std.array.at 0 == "/" then
+      'Error { message = m%"Absolute Path: Expecting a relative path, got an absolute path starting with "/"."% }
+    else
+      'Ok
+  )
+in
+let NoTraversal =
+  std.contract.from_validator (fun x =>
+    if x |> std.string.split "/" |> std.array.any (fun part => part == "..") then
+      'Error { message = m%"Invalid Traversal: Path contains ".."."% }
+    else
+      'Ok
+  )
+in
+let
+# A ValidTarget eliminates these pathological targets:
+# - Empty target, i.e., "",
+# - Absolute target, e.g., `"/etc/passwd"`,
+# - Parent traversals, e.g,. `"../../../../../../../etc/passwd"`.
+ValidTarget = std.contract.all_of [std.string.NonEmpty, RelativePath, NoTraversal]
+in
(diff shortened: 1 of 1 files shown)


lib/files.ncl
```diff
@@ -1,11 +1,22 @@
 let nix = import "./nix-interop/nix.ncl" in
+let RelativePath =
+  std.contract.from_predicate (fun x => !(x |> std.string.characters |> std.array.at 0 == "/"))
```
yannham, line 3: Nitpick: I would break the line to make it easier to read.

```suggestion
  std.contract.from_predicate (fun x => !(x |> std.string.characters |> std.array.first == "/"))
```

lib/files.ncl
```diff
@@ -1,11 +1,22 @@
...
+in
+let NoTraversal =
+  std.contract.from_predicate
+    (
+      fun x => !(x |> std.string.split "/" |> std.array.any (fun part => part == ".."))
```
yannham, line 8:
```suggestion
      fun x => x |> std.string.split "/" |> std.array.all ((!=) "..")
```

lib/files.ncl
```diff
@@ -72,7 +83,12 @@ let regenerate_files | Files -> nix.derivation.Derivation
           }
             file_descr.materialisation_method
         in
-        nix-s%"regenerate_function "%{copy_command}" "%{file_descr.file}" "%{file_descr.target}""%
+        let shell_escape = fun path =>
+          path
+          |> std.string.replace "\\" "\\\\"
+          |> std.string.replace m%"""% m%"\""%
+        in
+        nix-s%"regenerate_function "%{copy_command}" "%{file_descr.file}" "%{shell_escape file_descr.target}""%
```
yannham, line 91: I wonder if we shouldn't use `'` instead of `"` to also avoid unexpected shell expansions (what if the file name is like `$(/path/to/reverse/sell)` ?). Then we would have to escape `'` instead of `"`.

```suggestion
        let shell_escape = fun path =>
          path
          |> std.string.replace "\\" "\\\\"
          |> std.string.replace "'" "\\'"
        in
        nix-s%"regenerate_function "%{copy_command}" '%{file_descr.file}' '%{shell_escape file_descr.target}'"%
```

lib/files.ncl
Chan-Siu-Man: Ah ha, nickel has function section!

lib/files.ncl
Chan-Siu-Man: I intend to use `$'…'` (dollar-sign with *single*-quotes), instead of `"…"` (*double*-quotes), to quote the escaped string. This handles `$` in the string to escape correctly.

lib/files.ncl
```diff
@@ -72,7 +92,12 @@ let regenerate_files | Files -> nix.derivation.Derivation
           }
             file_descr.materialisation_method
         in
-        nix-s%"regenerate_function "%{copy_command}" "%{file_descr.file}" "%{file_descr.target}""%
+        let shell_escape = fun path =>
+          path
+          |> std.string.replace "\\" "\\\\"
+          |> std.string.replace m%"'"% m%"\'"%
+        in
+        nix-s%"regenerate_function '%{copy_command}' '%{file_descr.file}' $'%{shell_escape file_descr.target}'"%
```
yannham, line 100: I _think_ we should use `'` instead of `$'`. Since the string is coming from Nickel, Nickel is already going to expand escape sequences: if you write `let file = 'foo\nbar'`, the `\n` will be replaced by a new line, being equivalent to

```console
$ echo 'foo
quote>bar'
foo
bar
```
 
Using `$'..'` will also expand escape sequences somehow at a second level, that is `foo\\nbar` will be expanded to `foo\nbar` by Nickel, and here using `$'..'` will make a difference. But I feel like this isn't very useful since you would use `foo\nbar` to begin with. It would also mean that to correctly represent the filename `foo\bar` (where nothing is interpreted), you now need to write `foo\\\\bar` (interpreted as `foo\\bar` by "the first level", that is Nickel, and then as `foo\bar` by the shell string `$'..'`).

```suggestion
        nix-s%"regenerate_function '%{copy_command}' '%{file_descr.file}' '%{shell_escape file_descr.target}'"%
```

lib/files.ncl
```diff
@@ -72,7 +92,12 @@ let regenerate_files | Files -> nix.derivation.Derivation
           }
             file_descr.materialisation_method
         in
-        nix-s%"regenerate_function "%{copy_command}" "%{file_descr.file}" "%{file_descr.target}""%
+        let shell_escape = fun path =>
+          path
+          |> std.string.replace "\\" "\\\\"
```
yannham, line 97:
```suggestion
```

And if we use `'..'` as suggested below, I also think we don't need to escape `\` (just the sequence `\'` which is the only one to have a special meaning).

Chan-Siu-Man: Thanks for the suggestion. As the new PR uses `'…'` (single quotes) as suggested, there is no need to escape `\` (backslash).

lib/files.ncl
Chan-Siu-Man: Your argument is convincing. I have updated my PR to use `'…'` (single quotes) instead of `$'…'` (dollar-sign and single quotes) to quote commands.

Chan-Siu-Man: Ah, I take it back.

Upon testing, e.g.,, using `files."'".content = "…"` in an organist `project.ncl`, it appears that we need to use `$'…'` (dollar-sign and single quotes) instead of `'…'` (single quotes). For otherwise in  `'…'` (single quotes), we may need to expand every `'` (single quote) as `'"'"'` to close the single quote, use double-quotes to quote another single quote, then reopen the single quote.

Also, the generated script `regenerate-files` needs to make shellcheck happy, and for filenames with dollar-sign, e.g., `files."$pwd".content = "…"`, using `$'…'` (dollar-sign and single quotes) makes shellcheck happy but not `'…'` (single quotes), unless we suppress [SC2016](https://www.shellcheck.net/wiki/SC2016).

Chan-Siu-Man: Yes, you are right that shellcheck handles single-quotes (`'…'`) and dollar-sign and single-quotes (`$'…'`) differently.

Given this difference, if we want to use single-quotes (`'…'`), we need to suppress [SC2016](https://www.shellcheck.net/wiki/SC2016)—for example, to output an extra line `# shellcheck disable=SC2016` above every `regenerate_function` in `regenerate_one`. (Otherwise, shellcheck blocks the build of `regenerate-files` when there are dollar-signs `$` in filenames, such as `files."$one".content = "…"` in `project.ncl`.) 

I am fine with using either single-quotes (`'…'`) or dollar-sign and single-quotes (`$'…'`). Due to SC2016, I currently slightly prefer dollar-sign and single-quotes (`$'…'`).

yannham: Fair enough. I still feel like using `$` is a indirect work-around schell check bad behavior but it's relatively easy to change afterward anyway.

yannham (approved):
LGTM, given the conclusion of `$'...'` vs `'...'` discussion. Thanks!

